PR #4, opened 2017-08-25
Attachment shelve (local avatar entities)

thoys:
Attachments can now be grabbed without the need of rez-permissions.


thoys:
Removed the ESS way of creation as that was not very reliable. Moved to avatar entities. Feels way better.

misslivirose:
Is this live in dev-content to test?

thoys:
that is correct @misslivirose

misslivirose:
Looks good to me - QA Pass

PR #8, opened 2017-09-05
Fix bowling

misslivirose:
First pass on making bowling game work without rez rights. Server script handles spawning pins and bowling balls and is on a reset timer to prevent spamming.  Client script passes over the pin joint position, ball spawning position, and pin rotation.

There might be a race condition in setting up the pins. Note in testing. Test plan incoming.


misslivirose:
# Test Plan - Requires Hand Controllers, Multi-User
One person should have rez rights in the testing domain, the other should not have rez rights

1) Spawn bowling by going to Import Entities from URL and paste in the following: https://hifi-content.s3.amazonaws.com/liv/dev/bowling/bowlingAlley.svo.json **Test 1: Elements spawn**

2)  Setup/reset the pins and bowling balls by near-triggering in the button on the console (your hand must almost be inside of the button - you'll feel a haptic pulse). **Test 2: Both users with rez rights and without rez rights are able to set up their bowling alley**

3) Report back any weird findings related to: where and how the pins/bowling balls spawn, if there is strange behavior with multiple users trying out the bowling game

Nex-Pro:
testing this.

Nex-Pro:
### Users without rez rights can use the game. **SUCCESS**

**Bowling game 2 players issues seen:**

Physics are poor in terms of the expectations of bowlers. The ball does not seem to have enough inertia and tends to get thrown off at large angle at first contact rather than plough through the pins.

Downed pins can enter some kind of strange dancing state where they never settle down.

Occasionally pins do not reset despite the players hearing the reset sound when the button is pressed.

Sometimes the bowler sees that their ball has vanished but third party sees the ball near the other avatar's hand. If the third party touches the ball it drops to the ground and the avatar that was holding it can then see it again.

Many balls can be delivered as they are not cleaned up if the pins are re-spawned via the button.

Having the balls enter timeout mode after laser or grab is a poor design decision often times a user will move a ball before selecting the other. This can mean that before the user bowls the second ball it is removed.

The ball can get deflected towards the pins from the gutter due to the lane model having some design issues.

Bowler and observer may see different results following the pins being struck by the ball.

The release of the ball is sometimes delayed resulting in a bad shot.

Can't use the bowling game in a location such as maker where there are restrictions on entity size.

misslivirose:
Thanks @Nex-Pro! I think some of those are resulting of the new architecture and some were original design of the game. I'll iterate on this.

misslivirose:
Closing this due to age

PR #21, opened 2017-09-28
Highlighting

RebeccaStankus:


RebeccaStankus:
TEST PLAN FOR HIGHLIGHTING

This should be tested on an item with "attachmentItemScript.js" on it. These items will be in the store in empty or temporarily in dev-content.

-Near grab an attachment item
  It should have a yellow highlight around it while the avatar is holding it with near grab.
  The highlight should disappear when the item is released.

DomainContent/AvatarStore/Highlighting/highlight.js
```diff
@@ -0,0 +1,45 @@
...
+    _this.releaseGrab = function(entityID){
+        if (prevID !== 0) {
+            Selection.removeFromSelectedItemsList("contextOverlayHighlightList", listType, prevID);
+            prevID = 0;
+        }
+    };
+
+
+    var cleanup = function(){
+        Entities.findEntities(MyAvatar.position, 1000).forEach(function(entity) {
```
misslivirose: Why do you need to call this in find entities?

misslivirose:
Tested in dev-content with rawgit

PR #39, opened 2017-11-07
undo empty server script workaround for wearable children

dback2:
delete empty server script empty.js and remove adding it as server script to invisible child wearable entities in wearableServer.js since https://github.com/highfidelity/hifi/pull/11724 now fixes that


misslivirose:
@dback2 does this need QA? I don't think it does -- but want to confirm.

dback2:
Yes it does but we probably need to hold this for the next RC I guess? Or in the meantime they can test with the PR build? I can add a test plan tomorrow.

misslivirose:
@dback2 The new wearable scripts that use callEntityClient/callEntityServer method will remove the need for wearables to have any parents, but that won't get merged in for another few days so I'm fine if we want to test against the PR build for 11724 in the interim

Nex-Pro:
Testing

Nex-Pro:
Using 7413 and Win10 with vive

Nex-Pro:
### PASS

Result in testrail

Ran this in 11274

**Feedback:**

Good things: The headphones put no pressure on my ears at all ! They are lighter then a feather.
Bad things: The scarf failed miserably to keep out the cold.

PR #51, opened 2017-11-20
 WL 21588 - Playground with many blocks

cain-kilgore:
playground

Attached is a models.json.gz file, this is the playground!
Go in-game (preferably in HMD) and make something cool.

### Requirements
1. Before you launch High Fidelity Sandbox & Interface, copy [models.json.gz](https://github.com/cain-kilgore/hifi-content/blob/16db4a751b4aeee9570c5df170a52037bbd75bbe/Marketplace/Playground/models.json.gz) to `%appdata%\High Fidelity\assignment-client\entities`
2. Preferably HMD so you can play with the blocks.

### Test Plan
1. Confirm that your Sandbox domain is now a Playground and has lots of blocks (just over 400).
2. Confirm that you see a different variety of  blocks and colo(u)rs.
3. Confirm that the collision hull of each block is accurate and correct.
4. Confirm that when you leave and re-join, the blocks have stayed in the same place. (You can build something tall to confirm this)
5. Confirm that the block scale is big enough that you can walk directly under the Arch in the default Avatar/Scale.

### Note: If you laser the cylinders, they will be put back to static. This is a bug in HiFi and not related to this PR.


misslivirose:
@themelissabrown to follow up with how to distribute WL .json.gz files so we aren't distributing zipped files in the repo

PR #64, opened 2017-12-05
auto-destroy wearables if placed back on parent

dback2:
Test Plan:
-Use the monster beanie in dev-content A&J (it has the raw git scripts from this PR)
-Grab a wearable and immediately let go of it next to it's parent (should destroy)
-Grab a wearable and place it somewhere else away from the avatar and parent (should not destroy), and then grab it again and place it back on the parent (should destroy)
-Grab a wearable and place it on your avatar, then grab it again and place it back on the parent (should destroy)
-Grab a wearable and place it on your avatar, then have another Interface client log in that wasn't in world when you originally grabbed the wearable, and have that client's avatar grab it and place it back on the parent (should destroy)

### DIFF
--- a/DomainContent/AvatarStore/attachmentItemScript.js
+++ b/DomainContent/AvatarStore/attachmentItemScript.js
@@ -25,6 +25,7 @@
     var EMPTY_PARENT_ID = "{00000000-0000-0000-0000-000000000000}";
     var ATTACH_SCALE = 3;
     var MESSAGE_CHANNEL_BASE = "AvatarStoreObject";
+    var NOT_ATTACHED_DESTROY_RADIUS = 0.1;
   
     var messageChannel;
     var highlightConfig = Render.getConfig("UpdateScene.HighlightStageSetup");
@@ -37,6 +38,7 @@
     var prevID = 0;
     var listType = "entity";
     var attachDistance;
+    var initialParentPosition;
 
 
     var attachFunction = function() {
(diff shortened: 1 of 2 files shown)


misslivirose:
Tested in Dev-Content
- Replaced WearableServer.js and DesktopAttacher.js scripts on the purple monsta beanie
- Removed an item 
- Put item on head
- Removed item from head and placed it back on the shelf
- Item disappeared

misslivirose:
Had a thought -- we probably want to wrap the parent position in a check for nested type. Otherwise, I think that the parent position will be stored even if it's attached to an avatar

dback2:
Parent check added in preload before setting initial parent position

DomainContent/AvatarStore/attachmentItemScript.js
```diff
@@ -232,9 +241,16 @@
...
                         });
                     }
+                    isAttached = false;
                     Controller.triggerHapticPulse(TRIGGER_INTENSITY, TRIGGER_TIME, hand);
                 }
             }
+            
+            // If placed back within NOT_ATTACHED_DESTROY_RADIUS of the original parent entity 
+            // and it is not attached then destroy it (if the user is putting it back)
+            if (!isAttached && initialParentPositionSet && Vec3.distance(initialParentPosition, position) < NOT_ATTACHED_DESTROY_RADIUS) {
```
misslivirose: line length <128 otherwise looks good!

PR #66, opened 2017-12-05
Attachment joint check

RebeccaStankus:
Outlines to clarify the checkout process have been added. They will show whether an item is about to be deleted or scanned and link the overlay to it's matching entity. If an avatar does not have the joint that an item attaches to, the item will highlight in a special color to show that it cannot be attached.

TEST PLAN---
-Save 'outlineTesting.json' to your computer
-Go to Create Menu>Import Entities(.JSON) and select the outlineTesting.json file
-Change avatars to Robimo
-Pick up the top hat. The hat should highlight in red while you are holding it because Robimo does not have the joint that the hat attaches to
-Put on several gold necklaces. They should not highlight when you pick them up
-Enter the checkout area. Overlays should appear for the necklaces you are wearing
-Grab a necklace overlay and hold it in the recycle bin but do not release it. The overlay and it's matching entity should highlight in red to show that they will be deleted.
-Release the overlay. The overlay and matching entity should disappear
-Grab a necklace overlay and hold it in the scanner but do not release it. The overlay and it's matching entity should highlight in green to show that they will be scanned.
-Release the overlay. The overlay should reappear on the right side of the checkout stand.
-Grab a necklace that you are wearing. The matching overlay should highlight in blue.


misslivirose:
Looks awesome. Great work!

PR #67, opened 2017-12-08
tutorial update (attachments -> wearables rename)

thoys:


mukulHF:
I tested the PR, the spelling was changed in the tutorial screen according to the design spec. Thanks @thoys

PR #75, opened 2018-01-11
Turret PR Instructions

Delanir:
A turret system which shoots at player with bullets & effects.

## Loading Turret script
The turret script is located in Marketplace/turret/turret.js
Start the script by going to Edit -> Running Scripts -> FROM DISK -> Marketplace/turret/turret.js
## Using Turret script
The turret will spawn in front of your avatar and will start firing bullets. 

If enabled = true, will try to shoot at the player if player is within activeDistance units. It rotates to point at the player at the speed rotateSpeed. It begins firing shots at an interval, shootingInterval. If alwaysShoot is false, it waits until it has a clean shot of the player. It plays a one shot audio clip called chargeSound. As soon as chargeSound finishes, plays shotSound clip, rezzes shotParticle effect and rezzes/shoots bullet with bulletVelocity and bulletGravity parameters. Bullet plays flightSound as it whizzes thru the air. If smartBullet is true, it will attempt to follow the player. Bullet has particleEffect called flightParticle which trails behind it. If bullet hits something, it plays hitSound and rezzes hitParticle. When the bullet collides with something, it serializes a JSON object with this information and sends it to the Message system.


misslivirose:
According to Caitlyn, this PR is going directly into a specific domain and doesn't require ongoing maintenance. I will plan to close this if there are no additional requirements for maintaining from the content team

PR #76, opened 2018-01-11
Laser pointer

RebeccaStankus:
TEST PLAN
- Run spawnLaser.js. A laser should appear in front of your avatar
- Create a block, only changing the color property to black
- Equip the laser in either hand
- Turn the laser on using the trigger of the hand equipping the laser. The tip of the laser should light up and if there is an item with a collision hull in it's path, a red dot will light up on the intersected item.
- Aim the laser at the block. The red dot should appear on the block and should move as you move the laser pointer in your hand.
- Turn the laser off using the trigger of the hand equipping the laser. The tip of the laser should no longer be lit up and if there is an item with a collision hull in it's path, the red dot will disappear.
- Turn the laser on again and unequip it. The laser should automatically turn off when released.


YuppyPlays:
Testing.

misslivirose:
@RebeccaStankus is the issue that @YuppyPlays reports the limitation of the collision shape types that we've seen?

RebeccaStankus:
Yes @misslivirose I believe it was expected that pointing the laser at the sky will not produce a beam due to collision shapes.

misslivirose:
@YuppyPlays - I think that your feedback is definitely valid, but one that we might not be able to work around given some physics stuff on the engine side. We have a few internal cases around this and a specific domain content set we're using it for, so I'm going go ahead and merge, but keep the manuscript cases open so we can track them as the engine limitations are fixed up! Thank you!

PR #79, opened 2018-01-22
Fix checkout qml

misslivirose:
Test plan: 
(Stable) Join 'AvatarIsland' domain
Go to checkout station in 'Regalia' - second from the left (HMD)
Tablet should appear
Ensure it doesn't show empty blue screen but instead shows one of the following (dependently on whether wallet was configured or not, you are logged in or not etc. ):

- prompt to set up wallet
- purple check out instruction page

### DIFF
--- a/DomainContent/AvatarStore/CheckoutZone/checkoutZone.js
+++ b/DomainContent/AvatarStore/CheckoutZone/checkoutZone.js
@@ -29,7 +29,7 @@
     
     var TABLET = Tablet.getTablet("com.highfidelity.interface.tablet.system");
     var TABLET_ROTATIONAL_OFFSET = { x: 10, y: 220, z: 0 };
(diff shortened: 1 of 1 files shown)


elisa-lj11:
Test plan: **PASSED**
(Stable) Join 'AvatarIsland' domain
Go to checkout station in 'Regalia' - second from the left (HMD)
Tablet should appear
Ensure it doesn't show empty blue screen but instead shows one of the following (dependently on whether wallet was configured or not, you are logged in or not etc. ):

prompt to set up wallet 
purple check out instruction page

**I tried this while logged in, didn't get any blue screens. I logged out and walked up, and the first time I did I just saw the tablet home screen, but when I walked away and back again, I saw a prompt to log in, so not sure if that bug is related to this fix but otherwise I'd say it passed **

PR #85, opened 2018-02-02
Guitar

RebeccaStankus:
Test Plan
https://highfidelity.testrail.net/index.php?/cases/view/12741
https://highfidelity.testrail.net/index.php?/cases/view/12744


Marketplace/guitar/guitar.js
```diff
@@ -0,0 +1,666 @@
...
+                    }
+                }
+            }
+        },
+        setNeckColor: function(buttonID, params) {
+            neckGrabColor = params[0];
+        },
+        startNearGrab: function() {
+            if (!colorsShowing) {
+                this.createFingertipEntity("LeftHandMiddle4");
```
misslivirose: This should be skipped when the user is in desktop mode. Because it's manually being called here from clickReleaseOnEntity, we're getting these even when the user isn't using hand controllers in VR.

Marketplace/guitar/guitar.js
```diff
@@ -0,0 +1,666 @@
...
+                this.createFingertipEntity("RightHandMiddle4");
+                if (interval) {
```
misslivirose: What is the purpose of manually updating the fingertip entities in the updatePositions function? If they're attached to the avatar and don't have any kind of collisions with the avatar and full damping, I _believe_ we'd expect them to remain in place.

Marketplace/guitar/guitar.js
```diff
@@ -0,0 +1,666 @@
...
+                    Script.clearInterval(interval);
+                }
+                interval = Script.setInterval(this.updatePositions, POSITION_CHECK_INTERVAL_MS);
+                Script.setTimeout(function () {
+                    Script.clearInterval(interval);
+                }, POSITION_CHECK_TIMEOUT_MS);
+                this.createBodyOverlays();
+                this.createNeckOverlays();
+                numberHandsGrabbing = ONE_HAND;
```
misslivirose: I think trying to predict this is going to end up in strange states - especially because you're not technically "grabbing" with the second hand when you're playing.

Marketplace/guitar/guitar.js
```diff
@@ -0,0 +1,666 @@
...
+                    timeout = null;
+                }
+                timeout = Script.setTimeout(function(){
+                    _this.releaseGrab();
+                    timeout = null;
+                }, STILL_PLAYING_TIMEOUT_MS);
+                if (ampCheckInterval) {
+                    Script.clearTimeout(ampCheckInterval);
+                }
+                ampCheckInterval = Script.setInterval(this.updateAmpPositions, UPDATE_POSITION_MS);
```
misslivirose: I would move the amp check into startGrab but not set it as an interval - I think we should assume that the amps will not be moving during play, so it would be fine to update the position at the start of grabbing the guitar instead of on an interval.

Marketplace/guitar/guitar.js
RebeccaStankus: The problem was that, due to the avatar moving while the initial fingertip positions are calculated, the entities would not be in the correct place at the end of each finger at first, so I updated the position every 100ms for 3 seconds total. I have not seen any problems with the position after this process, but maybe the checks can occur less often or over a shorter period?

misslivirose:
Up on marketplace, passed marketplace QA

PR #86, opened 2018-02-04
Update camera app to use new protocolSignature API function

ctrlaltdavid:
Use Window.protocolSignature() in place of the deprecated location.protocolVersion().


ctrlaltdavid:
QA: Test that the Camera app (app-camera-move.js) works the same as before.

misslivirose:
Since this is an unpublished script, I went ahead and did a test pass on it internally. Seems clear to me.

PR #100, opened 2018-02-27
Updated support.js with blinking alert icon

FlameSoulis:
This update adds a blinking red circle when messages have been sent in support as well as putting it in position 12 on the tablet for easier locating by new users.


FlameSoulis:
Of course! Sorry about that.

When running the app, have one or two other users nearby, also running the app.

Have 1 user send a message with the app. This should, in turn for all OTHER users who have not opened the app, result in the blinking red circle appear on the icon of the app. Upon opening the app and closing it, the blinking should stop and the circle should not be visible.

Furthermore, the icon should, in a default loadout, appear in the lower right hand corner. This might have changed since the addition of the emotes button, so if this is the case, please let me know and the position can be addressed right away.

misslivirose:
Tested! The icon appears in the correct location, but only if there aren't additional extra scripts running. I think that this is fine for our first time user cases, since they won't likely be running additional scripts.

Internal Pass, I'll merge & update the version deployed in start!

PR #104, opened 2018-03-12
Wear app fixes for Fogbugz case: 11376

RebeccaStankus:


misslivirose:
@RebeccaStankus Can you link to the version of the URL we should use to test?

misslivirose:
Tested, looks good to me!

PR #115, opened 2018-04-26
Hub: flamethrower

dback2:


misslivirose:
Internally tested, approving for the functionality. We have known bugs with the endpoint based on engine raycast distance calculation.

PR #117, opened 2018-04-30
Hub domain: Plates

misslivirose:
**Test Plan - Requires HMD; test without rez rights**
 
- Go to the hub into the 'Flying Saucer'
- Pick up a plate from the end of the table
- Throw the plate aggressively against something else
- The plate should shatter
- Pick up a piece of the broken plate
- Throw it aggressively against something else
- The piece should shatter and play a particle effect


elisa-lj11:
I was not able to pick up the plate.

elisa-lj11:
I found the grabbable plates and they worked great.

PR #125, opened 2018-05-04
Hub: Flamethrower Shape Ray Cast Fix

dback2:
plus a couple other smaller fixes for the desktop overlay and to disable the not fully debugged server-side end fire sounds


misslivirose:
@dback2 is the previous JSON link good to test or is there a separate one we should use for this PR?

dback2:
@misslivirose I've updated https://hifi-content.s3.amazonaws.com/davidback/development/flamethrower/flamethrowerGitTest.json to target the new scripts in this PR

misslivirose:
Internally QA tested new JSON. Fires appear at ends of rays. The handshake statue is a bit wonky but everything else behaves as expected.

PR #126, opened 2018-05-04
moved case-sensitive lowercase folders over to their correctly cased folders

thoys:
domaincontent -> DomainContent
marketplace -> Marketplace


thoys:
no QA required

PR #128, opened 2018-05-07
Hub: Flamethrower - Normalized ray casts for all

dback2:
Remove the shape-specific ray cast and use normalized ray casts for both entities and avatars

https://hifi-content.s3.amazonaws.com/davidback/development/flamethrower/flamethrowerGitTest.json points to latest script changes here


misslivirose:
Tested internally, passes QA

PR #138, opened 2018-05-11
Move Shapes app from hifi to hifi-content

ctrlaltdavid:


ctrlaltdavid:
QA not required because no code changes.

PR #161, opened 2018-06-21
Pucks sliding quick fix

robin-k-wilson:
Utils script that, when running, enables users to use the default walk animations while in full body pucks. 

Testing:
- Load script via URL in Interface: https://hifi-production.s3.amazonaws.com/robin/bug-fixes/trackers/pucks-sliding-fix1.js
- Without pucks: The avatar is unaffected.
- In pucks: 
    - Avatar should only be affected when holding and briefly after holding down the moment keys.
    - Holding down movement key, the avatar will smoothly transition into the default walk animation.
    - After letting go, the avatar will smoothly sync hands and feet back to the puck positions.


Utilities/Trackers/pucks-sliding-fix1.js
```diff
@@ -0,0 +1,131 @@
+/*
```
misslivirose: Please use the standard header comment formatting here

Utilities/Trackers/pucks-sliding-fix1.js
```diff
@@ -0,0 +1,131 @@
...
+    function update(dt) {
+
+        var areHipsInPucks = Controller.getPoseValue(Controller.Standard.Hips).valid;
+
+        if (areHipsInPucks) {
+
+            var hasActiveDriveKey = false;
+
+            for (var i in OVERRIDDEN_DRIVE_KEYS) {
+                if (MyAvatar.getRawDriveKey(OVERRIDDEN_DRIVE_KEYS[i]) !== 0.0) {
```
misslivirose: Is there a reason that this 0.0 is different than using a non-float 0?

Utilities/Trackers/pucks-sliding-fix1.js
```diff
@@ -0,0 +1,131 @@
...
+                // disable mappings
+                disable(dt);
+            }
+        }
+    }
+
+    setUp();
+    Script.update.connect(update);
+
+    Script.scriptEnding.connect(function () {
```
misslivirose: Should the script also disconnect update here?

PR #163, opened 2018-06-26
stage manager app

RebeccaStankus:
FEATURES

- Raise and lower individual web screens on left and right of stage
- Raise and lower both web screens simultaneously
- Set up web screens with embedded slides link
- Navigate through slides
- Reset slides
- Raise and lower backdrop
- Silence everyone except Philip and Jessica, and 4 ambisonic injectors by name
- Unsilence the above
- Play 2 different audio clips 
- Adjust volume on the above audio clips
- Change set background
- Restore set background


DomainContent/FaceToFace/StageManager/enterStageZone.js
```diff
@@ -0,0 +1,115 @@
...
+                x: -0.0007781982421875,
+                y: 0.9099717140197754,
+                z: 0.4146333932876587
+            },
+            position: {
+                x: 99.6183,
+                y: -0.6548,
+                z: 32.4620
+            },
+            url: "https://hifi-slides.glitch.me/",
```
misslivirose: This is fine for now, but we might end up hitting some weirdness if we end up putting an entity whitelist up and miss that this is on glitch instead of our S3 bucket.

DomainContent/FaceToFace/StageManager/stageManager.js
```diff
@@ -0,0 +1,208 @@
...
+    function onClicked() { 
+        if (open) {
+            tablet.gotoHomeScreen();
+        } else {
+            tablet.gotoWebScreen(appPage);
+        }
+    }
+
+    function onWebEventReceived(event) {
+        if (typeof event === 'string') {
```
misslivirose: This would be easier to read as a switch statement - also, ordered by most commonly used (I'd imagine that both screens might be more common than one or the other?)

DomainContent/FaceToFace/StageManager/stageManager.js
```diff
@@ -0,0 +1,208 @@
...
+            name: "Mute All",
+            script: "http://hifi-content.s3.amazonaws.com/rebecca/Today/muteTest.js?125",
+            type: "Sphere",
+            userData: "{\"grabbableKey\":{\"grabbable\":false}}",
+            visible: false
+        });
+        muteBalls.push(muteBall);
+    }
+
+    function sound() {
```
misslivirose: can this function be named something more descriptive?

DomainContent/FaceToFace/StageManager/todayButton.js
```diff
@@ -0,0 +1,155 @@
+//
+// todayButton.js
```
misslivirose: Is this still needed since we have the app which handles these functions?

PR #164, opened 2018-06-26
Bell for jessica

RebeccaStankus:
TEST PLAN
- Equip the bell in HMD via lower trigger
- Pull the top trigger to make the bell ring
- Unequip the bell with lower trigger


YuppyPlays:
Testing.

YuppyPlays:
**Testing complete:**

Equip the bell in HMD via lower trigger **Check**
Pull the top trigger to make the bell ring **Check**
Unequip the bell with lower trigger **Check**


**Notes:** I assume Lower trigger refers to grip on a Vive, which works as expected.
The grips can be used to equip or unequip. trigger will equip and then work as expected to ring the bell.
Holding the trigger down makes the bell go extremely fast, which I quite like. Volume is also good.

**Pass**

PR #168, opened 2018-07-06
Improve the Laser Pointer

zfox23:
Three main improvements here:
1. Make the on/off sound come from the laser pointer instead of `MyAvatar.position`
2. If there's no `intersection`, make the laser pointer go off to "infinity" (actually just 100m) instead of not changing the previous pointer
3. Fix the laser pointer beam appearing to come from behind the pointer model by using `registrationPoint`

### DIFF
--- a/Marketplace/Laser Pointer/laser.js
+++ b/Marketplace/Laser Pointer/laser.js
@@ -131,10 +131,24 @@
                     injector.stop();
                 }
                 injector = Audio.playSound(sound, {
-                    position: MyAvatar.position,
+                    position: Entities.getEntityProperties(_this.entityID, ['position']).position,
                     volume: AUDIO_VOLUME_LEVEL
                 });
             }
+        },
+        
+        addFocus: function(size, position) {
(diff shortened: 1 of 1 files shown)


zfox23:
@RebeccaStankus, could you please take a look at this?

Marketplace/Laser Pointer/laser.js
```diff
@@ -147,88 +161,90 @@
                 direction: beamDirectionNormalized
             });
             var entityIntersection = Entities.findRayIntersection(beamPickRay, true, [], doNotRayPick);
-            var entityIntersectionDistance = entityIntersection.intersects ? entityIntersection.distance : Number.MAX_VALUE;
+            var entityIntersectionDistance = entityIntersection.intersects ? entityIntersection.distance : 100;
```
RebeccaStankus: Can we name these 2 magic numbers?

zfox23: Sure! sec

PR #171, opened 2018-07-10
avatar counter updates by query via domain API

RebeccaStankus:
This counter is currently set to check the number of avatars in Zaru. We can change the domain it counts by changing the URL in the request.
TEST PLAN
- add server script, "avatarCounter.js" to a text entity in zaru
- it should update with the current number of avatars in the domain and update every second


DomainContent/LoadTest/AvatarCounter/avatarCounter.js
```diff
@@ -0,0 +1,45 @@
...
+
+    AvatarCounter.prototype = {
+        interval: null,
+
+        preload: function(entityID) {
+            _this.entityID = entityID;
+            _this.interval = Script.setInterval(function() {
+                var userData = JSON.parse(Entities.getEntityProperties(_this.entityID, 'userData').userData);
+                var requestURL = BEGIN_DOMAIN_API_URL + userData.domainAPIURL;
+                request(requestURL, function (error, data) {
```
misslivirose: It might be good to confirm that the userdata contains the domain API URL before requesting this

DomainContent/LoadTest/AvatarCounter/avatarCounter.js
```diff
@@ -0,0 +1,45 @@
...
+
+        unload: function() {
+            if (_this.interval) {
+                Script.clearInterval(_this.interval);
+            }
+        }
+    };
+        
+    return new AvatarCounter;
+});
```
misslivirose: Newlines

misslivirose:
Works in hifi://staging3 - internally tested and approved

PR #175, opened 2018-07-17
Korea lock instruments

robin-k-wilson:
In a room with korea instruments, the guide can swipe a hand over "the master lantern" to unlock instruments in the room. The "instrument lanterns" will light up when the music instruments are grabbable and be off when they are locked, mirroring the lit state of the master lantern.

**Setup and Normal Use Case**
1. Go to "hifi://korea"
2. To to the Music room. The "master lantern" is next to the "MUSIC" sign in the back of the room
3. Ensure you're on the script hard-coded white list **_-- please comment on this PR with your hifi username if you need to be added to the whitelist for the Korea domain AND/OR the script_**
4. Check the instruments:
    - If the lights around the room are off, check that the instruments are locked.
    - If lights are on, check that instruments are grabbable and play music.
6. Swipe over the "master lantern," the lights will switch states from on to off or from off to on, the lanterns near the instruments will mirror the master lantern.
7. Check the instruments:
    - If the lights around the room are off, check that the instruments are locked.
    - If lights are on, check that instruments are grabbable and play music.

**Specific Cases to ensure are always true**
- Instrument lights always match the master lantern state.
- All music making Instruments are:
            - Locked when lights are off
            - Grabbable when lights are on
- Only people on the instrument whitelist are able to toggle the master lantern

### DIFF
--- a/Prototyping/Domains/Korea/InstrumentSwitch/instrumentSwitch.js
+++ b/Prototyping/Domains/Korea/InstrumentSwitch/instrumentSwitch.js
@@ -14,7 +14,7 @@
 
 (function () {
 
-    var ALLOWED_USERS = ['Firebird25', 'Alan_', 'jyoum', 'tinydancer', 'MissLiviRose'];
+    var ALLOWED_USERS = ["Firebird25", "Alan_", "jyoum", "tinydancer", "MissLiviRose"];
 
     var isAllowedUser = false;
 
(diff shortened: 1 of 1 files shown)


AndySeattle:
Just finished testing this; behaving as expected with the Master light, Looks good! 

As long as it is intended that we can pick up the "non playing" instruments for inspection on the table by the light, i am good to go here. (Talking to Robin, i believe that is the expected behavior; that we are able to inspect the instruments that can not play music, then when we switch on the light, we can pick up the ones that can.)

PR #177, opened 2018-07-24
Fixed particle changes with latest build

RebeccaStankus:
Test Plan
-Go to "theSpot" and use the material gun.
- Upon firing, the particles should be small like spraying paint from the end of the gun and where it hits something.


AndySeattle:
Took a look at this in VR mode, feels a bit understated to me, but it works. Just compared to the "rainbow thrower" this one feels to emit kind of a small effect. Attaching two gifs for comparison, but that is just an opinion. As far as the PR it looks good, we are throwing colors and changing the wall to the color particle throw. 
hifi-snap-by-andy_batman-on-2018-08-07_10-45-30

hifi-snap-by-andy_batman-on-2018-08-07_10-46-48

PR #179, opened 2018-07-25
Bouncer usernames

robin-k-wilson:
Zone that acts as a bouncer, can configure with usernames hardcoded inside file, usernames inside userData, and a specific marketplace item id.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -69,7 +69,7 @@ module.exports = {
         "new-cap": ["error"],
         "no-console": "off",
         "no-floating-decimal": ["error"],
-        //"no-magic-numbers": ["error", { "ignore": [0, 1], "ignoreArrayIndexes": true }],
+        "no-magic-numbers": ["error", { "ignore": [0, 1], "ignoreArrayIndexes": true }],
         "no-multiple-empty-lines": ["error"],
         "no-multi-spaces": ["error"],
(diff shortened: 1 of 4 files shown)


.eslintrc.js
```diff
@@ -69,7 +69,7 @@ module.exports = {
         "new-cap": ["error"],
         "no-console": "off",
         "no-floating-decimal": ["error"],
-        //"no-magic-numbers": ["error", { "ignore": [0, 1], "ignoreArrayIndexes": true }],
```
MiladNazeri: Is this supposed to be in the pr?

Prototyping/Sit/sit.js
```diff
@@ -0,0 +1,720 @@
+//
```
MiladNazeri: I don't think this is meant to be in here

Prototyping/Sit/sitServer.js
```diff
@@ -0,0 +1,91 @@
+/* globals Entities, Script, AnimationCache, Settings, MyAvatar, DriveKeys, AvatarList,
```
MiladNazeri: I don't think this is meant to be here

MiladNazeri:
Bouncer script looks good.  Didn't see anything glaring.  
A few things I don't think were meant to go in this pr though.  Please confirm.

PR #186, opened 2018-08-24
Meowing Cat 

misslivirose:
This script relies on having cat models and animations in the domain asset browser. The basic code allows you to add meow.js to a cat model to play a meow sound when it's clicked on. 

**Test Plan**
1. Go to hifi://studio/-198.2,28.4,-81.6
2. Click on the small white cat that is sitting on the chair
3. Confirm that the kitten meows and plays a meowing animation 

Note: the animations used currently are in an unfinished state, but shouldn't block QA.


AndySeattle:
@misslivirose my only feedback would be to let the animation finish before another one can be triggered? as a symptom of the ability to spam animations, the cat can start to look a bit max hedrum, tweaking from animation to the next starting a different reaction in the middle of the previous. But if the purpose of this is to block out code of an intractable cat, i am a pass!

MiladNazeri:
quick thought, maybe add a userData option that makes the meow local only.  Thinking in terms of the cat cafe so people don't grief with meows.

misslivirose:
@MiladNazeri To clarify, this is not meant to go as-is into the Cat Cafe, but to provide a base script that we can customize once we have more definition around the project deliverables 🐱

DomainContent/FaceToFace/Cats/meow.js
```diff
@@ -0,0 +1,68 @@
...
+    var MeowEngine = function(){
+
+    };
+
+    MeowEngine.prototype = {
+        remotelyCallable : ['meow'],
+
+        preload : function(entityID) {
+            selfEntityID = entityID;
+            selfPosition = Entities.getEntityProperties(entityID, 'position').position;
```
MiladNazeri: not related, but wondering if we need to update the documentation to reflect that the second parameter can also be a single string.  Think we have it listed as only an array of strings.

DomainContent/FaceToFace/Cats/meow.js
```diff
@@ -0,0 +1,68 @@
...
+                Script.setTimeout(function(){ 
+                    Entities.editEntity(selfEntityID, {
+                        animation: {url: IDLE_ANIMATION_URL, loop: true, fps: 30}
+                    });
+                }, 1000);
+            }
+        }, 
+        
+        unload: function(){
+            Script.clearInterval(meowInterval);
```
MiladNazeri: meowInterval doesn't get used anymore it looks like(guessing you had interval first and changed to timeout)

DomainContent/FaceToFace/Cats/meow.js
```diff
@@ -0,0 +1,68 @@
...
+    var selfEntityID;
+    var selfPosition;
+
+    var MEOW_SOUND_URL = Script.resolvePath('./resources/sounds/meow.mp3');
+
+    var MEOW_ANIMATION_URL = 'atp:/kitten_Meow.fbx';
+    var IDLE_ANIMATION_URL = 'atp:/kitten_Idle.fbx';
+
+    var meowSound = SoundCache.getSound(MEOW_SOUND_URL);
+    var meowInterval;
```
MiladNazeri: doesn't get used

misslivirose:
@MiladNazeri - Mind re-reviewing? This is the version we used in the cafe, cleaned up. Doesn't require a new QA pass.

PR #188, opened 2018-08-30
Shower Dance Walk Fixes

MiladNazeri:
This fixes walking through the shower.  The forward walk was broken originally.  
I haven't been able to reproduce avatars being bunched up, so this doesn't address that if it is still an issue.

Test Plan:

- [x] Walk through the shower

- [x] Do you have a new walk animation that is the same but in reverse when you move the opposite direction?

- [x] Press the reset button

- [x] Do you walk normally?

- [x] Repeat the process several times and make sure there aren't any oddities.

### DIFF
--- a/DomainContent/Hub/Animation Shower/resetButton.js
+++ b/DomainContent/Hub/Animation Shower/resetButton.js
@@ -11,7 +11,7 @@
 (function() {
     var SOUND_URL = Script.resolvePath("Sounds/undo-notif.wav");
 
-    var walkRoles = ["walkFwd", "idleToWalkFwd", "walkBwdNormal"];
+    var walkRoles = ["walkFwdNormal", "idleToWalkFwd", "walkBwdNormal"];
     var sound = SoundCache.getSound(SOUND_URL);
 
     function resetAnimations() {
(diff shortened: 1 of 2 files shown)


AndySeattle:
I dug into this more, it looks like the mirror script is not loading for "me" on the first try in help domain so it was a coincidence, BUT we are still carrying over the walking effect to other domains and not resetting it on leaving the spot. Is this international @MiladNazeri (I will follow up the mirror not loading separately. but for the morbidly curious, here is what im seeing on loading into help the first time: 

[09/04 13:42:12] [WARNING] [hifi.scriptengine] Error loading script from URL  QUrl("https://hifi-content.s3.amazonaws.com/rebecca/mirrorReflection.js") ( "Error" )

)

MiladNazeri:
I just tried going to that URL and I got an accessed denied.  

re: should it still be left on?  Um I'm not sure what the original intent was.  @misslivirose would definitely be a good source for that.  
We can either put it on a timer, or add a hook so if the domain changes your animation gets restored.  Shouldn't be too hard if so, we'll wait for Liv to chime in.

misslivirose:
Intent for the original design was that it persisted but it was back and forth on whether or not to change it. I'd suggest that we not aim to change functionality without tying it into a manuscript case as a suggestion that gets vetted against our other deliverables. 

As for the mirror in help, I think this was addressed in the slack thread but that's not one of our team domains and we should follow up with them separately about using a more recent version.

AndySeattle:
K. I am good then thanks @misslivirose it looks good outside of it "sticking" when you leave the domain. moving to Pass, and i believe the Mirror on Help has already been fixed at this point, but i will follow up with that team. thanks! marking as qa passed!

PR #194, opened 2018-09-19
Explosion, following eyes, and random interval sound emitter

RebeccaStankus:


DomainContent/Halloween/ANJExplosion/aNjExplosionServer.js
```diff
@@ -0,0 +1,45 @@
+(function () {
```
misslivirose: Add header comment block

DomainContent/Halloween/FollowingEyes/catEyes.js
```diff
@@ -0,0 +1,84 @@
...
+        leaveEntity: function() {
+            Overlays.deleteOverlay(rightEyeOverlay);
+            Overlays.deleteOverlay(leftEyeOverlay);
+            if (interval) {
+                Script.clearInterval(interval);
+            }
+        },
+
+        unload: function() {
+            _this.leaveEntity();
```
misslivirose: I would suggest that you also check if the overlays exist in `unload` and if they do, remove them. It would be an edge case that they didn't leave the entity before the script was unloaded but just in case, the cleanup would still work.

DomainContent/Halloween/FollowingEyes/forestEyes.js
```diff
@@ -0,0 +1,88 @@
...
+            eyes.forEach(function(eyeOverlay) {
+                Overlays.deleteOverlay(eyeOverlay);
+            });
+            if (interval) {
+                Script.clearInterval(interval);
+            }
+        },
+
+        unload: function() {
+            _this.leaveEntity();
```
misslivirose: Same behavior with the cat eyes, suggest checking for the overlays and deleting as a precaution in `unload` in addition to `leaveEntity`

DomainContent/Halloween/FollowingEyes/catEyes.js
RebeccaStankus: leaving in because it calls "leaveEntity" which checks for overlays

RebeccaStankus:
TEST PLAN
- Enter the domain "Zombies" and you should hear sounds from various places at random intervals such as a beeping SOS sound from with in the boat, radio emergency talking from the boat speaker, screams, an ambulance siren, ambulance radio static, and spooky ambient background noise in A&J and Regalia
- walk along the path to your right and there is a black cat sitting on a flower planter and as you approach, it's eyes appear/open and follow you as you walk by
- continue on the path around the back side of A&J
- After passing the cars on your left a loud explosion occurs
- You see a burst of flames with smoke and a mannequin flies out of the 2nd floor window
- After walking up the ramp ahead, the scene returns to normal behind you by A&J
- Go to the domain "Incipio" and approach the abandoned car that can be seen from the fire.
- As you get close to the car, several pairs of eyes should appear throughout the forest, looking at you
- As you walk toward them, they disappear

Sweet dreams!!!

PR #199, opened 2018-09-21
Whisper Utility

robin-k-wilson:
**Directional Whisper**
Setup: 
Create a box named "focusObj";
Run the script in your "Running scripts". Should always hear the "quick over here" sound towards the "focusObj". 
Refresh script in a different position. 

Can be <100 feet away and still hear the sound clearly in the correct direction.


YuppyPlays:
Testing complete:

All **Pass** :)

PR #203, opened 2018-10-11
Digital photo album with readme

misslivirose:
See Readme for complete details about this item.
Short testing plan:

- [x] Download the digital-photo-frame.json file from this PR and save it to your local machine
- [x] Import the object into Interface in a domain where you have rez rights
- [ ] Add custom photos to the userData and refresh the server script on the digital-photo-image entity
- [ ] Change the 'changeSpeed' variable in userData and refresh the server script on the digital-photo-image entity

Expected: you can customize the images and how quickly they change


AndySeattle:
Sent Salina's way to test in her new fancy office

salinacsoto:
Hello @misslivirose ! So while testing the digital photo frame I ran into the issue of when I went to change the user data the debug window brought up theses errors. I was unable to save any of the changes and the frame would relock on me every ten seconds. 
untitled

misslivirose:
@salinacsoto - Sorry about that, the version in the test file was referencing an old script. Can you please clear content and re-try? It shouldn't being automatically locking. The errors in the log are in edit, not the script.

salinacsoto:
@misslivirose - So I was able to edit the user data but I noticed two issues while testing. 

- First, when you change the number for changeSpeed for the photos, all of them but the first image (0) will reflect that change. The first image will be double the time of any number you put in changeSpeed.
- The second issue was that the 5th image never shows up. The Frame only flickers through four images instead of all five that is presented in the default user data.

misslivirose:
Thanks @salinacsoto - will address

Marketplace/photoFrame/digitalFrame.js
```diff
@@ -0,0 +1,67 @@
...
+    };
+
+    // The preload function is called when the script is added, reloaded, or the server starts up
+    this.preload = function(entityID) {
+        _entityID = entityID;
+        try {
+            var customProperties = JSON.parse(Entities.getEntityProperties(_entityID, 'userData').userData).digitalFrame;
+            photos = customProperties.photos;
+            speed = customProperties.changeSpeed * SECONDS_TO_MILLISECONDS;
+            Script.setInterval(changePhotos, speed);
```
robin-k-wilson: I tested this implementation of Script.setInterval and Script.clearInterval without saving the returned reference to the QTimer. I was unable to clear the interval in my client. Does the server script behave differently? 
I suggest saving the QTimer reference with a global variable intervalID to later use in unload's Script.clearInterval function. 

// As a global variable 
var intervalID = null;

// Replace line with:
intervalID = Script.setInterval(changePhotos, speed);

Marketplace/photoFrame/digitalFrame.js
```diff
@@ -0,0 +1,67 @@
...
+            print ("Error loading details from userdata - using defaults");
+            photos = DEFAULT_PHOTO_ARRAY;
+            speed = DEFAULT_TRANSITION_SPEED * SECONDS_TO_MILLISECONDS;
+            Script.setInterval(changePhotos, speed);
+        }
+    };
+
+    // The unload function is called when the script is removed, reloaded, or the server shuts down
+    this.unload = function() {
+        Script.clearInterval(changePhotos);
```
robin-k-wilson: Script.clearInterval(intervalID);
intervalID = null;

Marketplace/photoFrame/digitalFrame.js
misslivirose: As written, this still stops when the script is removed, but does through the incompatible type error. Not sure why it works, but will make the suggested change anyway.

PR #209, opened 2018-10-23
Fix lifetime on dishes and plates

MiladNazeri:
This is a fix for the lifetime on dishes and plates. 

I currently am using a value of 60 as a default for all things as I saw another mention of that for when a plate breaks.  

I'll set up a domain with updated scripts to test on.  Contact me when ready.  

- [ ] Go to a plate to break

- [ ] Clone it but try not to break it

- [ ] Does it go away after 60 seconds?

- [ ] Clone it again and smash it

- [ ] Are the particles that are created when a dish is broken smaller than the previous version?

- [ ] Does it go away after 60 seconds or earlier?

- [ ] Find a Tequila glass and clone it

- [ ] Does it go away after 60 seconds?

- [ ] Clone it again and smash it

- [ ] Does it go away after 60 seconds or earlier?

### DIFF
--- a/DomainContent/Hub/Space Juice/spaceJuiceSpawner.js
+++ b/DomainContent/Hub/Space Juice/spaceJuiceSpawner.js
@@ -69,8 +69,7 @@
                 serverScripts: Script.resolvePath("spaceJuiceServer.js"),
                 shapeType: "simple-hull",
                 type: "Model",
-                dynamic: false,
-                lifetime: LIFETIME
+                dynamic: false
             });
 
             Entities.addEntity({
@@ -92,8 +91,7 @@
                 shape: "Cone",
                 type: "Shape",
                 userData: "{\"grabbableKey\":{\"grabbable\":false}}",
-                script: Script.resolvePath("glassPrimitiveCatch.js"),
-                lifetime: LIFETIME
+                script: Script.resolvePath("glassPrimitiveCatch.js")
             });
         }
     };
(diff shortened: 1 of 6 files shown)


MiladNazeri:
Just realized "tequila shots" on the card probably meant space juice and not the tequila shots in Mexico.  I thought they were just repurposed.  Fixing.

salinacsoto:
--Go to a plate to break
--Clone it but try not to break it
--Does it go away after 60 seconds? - **Yes- PASS**
--Clone it again and smash it
--Are the particles that are created when a dish is broken smaller than the previous version? - **Yes- PASS**
--Does it go away after 60 seconds or earlier? - **Yes- PASS**
--Find a Tequila glass and clone it
--Does it go away after 60 seconds? - **Yes- PASS**
--Clone it again and smash it
--Does it go away after 60 seconds or earlier? - **Yes- PASS**

@AndySeattle You can mark this one as a PASS

AndySeattle:
Quick question before i mark as passed, @MiladNazeri @salinacsoto did we verify this just in Milad sandbox or have we verified this in the actual spot yet?

MiladNazeri:
We've only done it on my sandbox with an archived version of the spot.  I'll ask @misslivirose about the best approach to put them in there since they are removed currently and she is more familiar.

AndySeattle:
k. Salina passed these, looks good feel free to merge and place in the spot :)

salinacsoto:
One issue I came across with the plate spawner was when I moved a plate slightly next to where the spawner is that the spawner will not rez another one till the plate that is next to it is gone. Any other plate that was grabbed and moved a bit of ways away from the spawner would trigger the spawner to make another plate but with this case it was waiting for the one close to it to be gone. Is this by design?

misslivirose:
@salinacsoto - yes, by design, it will only spawn if there isn't one nearby - I think it's maybe .5 meter?

PR #213, opened 2018-10-29
Consolidated halloween content and korea content into proper directories

robin-k-wilson:


misslivirose:
No functional changes, can merge without QA

PR #214, opened 2018-10-31
Bingo

RebeccaStankus:
TEST PLAN--Requires at least 3 people, a boss and a user, and someone to try to get a card after registration closes.
-Bingo Boss runs     http://hifi-content.s3-us-west-1.amazonaws.com/rebecca/BongoBingo/bingoApp.js
-Boss presses "Lights on" and "open registration"
-User clicks "Play Bingo", hears a sound clip and then opens their tablet app "BINGO". A gem appears above their head. If they close the app or switch apps, the gem disappears.
-Boss clicks "Close Registration" and no one else can get a card.
-Boss clicks wheel to spin for a random number. When the wheel lands on a number, a sound goes off and the matching number on the wall lights up
-User checks off a number on their card if they have it by clicking and will hear a sound. They can unclick a number if they want and should hear a sound for that as well. The letters of the BINGO header each have a different hidden sound that can be heard by clicking the square.
-When a user has a line of squares across, down , or diagonally, they step up to the bingoscanner machine to validate and a spotlight shines on them and angelic sound plays. A card appears above the scanner for all to see and it shows which numbers they have and the ones that have been called are highlighted. The machine plays a thinking sound. If they truly have a line of numbers that have been called, the card removes all highlights except for the winning row and a winning sound plays and confetti appears. If they do not have a winning card, the loser sounds play. After leaving the bingoscanner, the  spotlight goes off and 3 seconds later, the card disappears.
-Boss can close the game or start a new round via the tablet.


DomainContent/BongoBingo/bingoWheel.js
```diff
@@ -0,0 +1,272 @@
...
+            });
+            Entities.editEntity(bingoNumberTexts[0], {
+                locked: true
+            });
+            if (_this.interval) {
+                Script.clearInterval(_this.interval);
+            }
+        },
+
+        mousePressOnEntity: function(entityID, mouseEvent) {
```
misslivirose: It might be worth investigating if the wheel works with this functionality offloaded to a server script, so that when you click on the wheel, if you have permissions, you request the server to handle this. I'm not sure whether the ESS can modify velocity properties successfully, but I think that would make the wheel more robust.

DomainContent/BongoBingo/bingoWheel.js
```diff
@@ -0,0 +1,272 @@
...
+                                    }, LIGHT_BLINK_INTERVAL);
+                                }
+                            }
+                        }, ANGULAR_VELOCITY_CHECK_MS);
+                    }, CHECKING_INTERVAL_DELAY);
+                }
+            }
+        },
+
+        lightOn: function(lightNumber) {
```
misslivirose: These should be moved to an entity server script that handles the editing to avoid the lock/unlocks

DomainContent/BongoBingo/card/card.js
```diff
@@ -0,0 +1,204 @@
+//
```
misslivirose: @robin-k-wilson - can you review this file since I wrote the original?

AndySeattle:
Still needs a code review, however, has passed a full test pass several times, and did really well when ran the first time publically. Holding off on marking as qa passed till this code review is done. Will run another test pass after requested changes have been made and reviewed.

PR #217, opened 2018-11-09
Add support for Mikomiko SIM joints and allow one joint threads

luiscuenca:
https://highfidelity.manuscript.com/f/cases/18765/Flow-js-doesn-t-support-single-bone-chains
This PR makes possible parsing sim joints from Mikomiko avatars and adds them as flow joints.
It also allows the simulation of threads with only one joint by attaching a dummy joint and adding some stiffness.


misslivirose:
Approved by platform team - merging

PR #233, opened 2018-12-12
Adding a mirror with a button as a Blueprint

misslivirose:
Test Plan:

1. Import the following marketplace entity by URL: http://mpassets.highfidelity.com/40b93ca0-c08f-42a8-9034-462f8bfa5fcb-v1/mirror_with_button.json

2. Confirm that the mirror appears, and turns on/off when the button is clicked (desktop) and far-triggered (HMD)

Note: there appears to be a bug that I'm looking into where the first several times you import items from marketplace via the 'Import Entities from URL' menu, the scripts do not run until you open the Create menu (or wait quite a long time for them to load). That issue is not specific to this piece of content, but if you hit it in testing, it will be a valuable info point.


salinacsoto:
Following strictly the test plan it passes.

Though when I tried resizing the mirror and clicking the button, I experienced a lockup and was unable to select anything on my computer. Thankfully high fidelity recovered after a few minutes but I just want to know if this is a known issue for resizing. If it is, just a warning attached to the mirror about making the mirror a bit too large will cause latency issues would be great to have.

The local dimensions in which caused me this issue of the mirror plane was X:7.7091 Y:4.8355 Z:0.0276

I also noticed the issue of only being able to have one mirror run at once. If I import two mirrors I have to make sure one of them is off in order to use the other one and vice versa no matter how far the mirrors are from each other in the domain. 

One thing that was odd while testing this in Studio was when I turn on the mirror my reflection was being *reflected* on wall materials near by and when I would shut off the mirror a still of me would take its place. 
hifi-snap-by-skelina-on-2018-12-12_13-59-38

misslivirose:
Only being able to use one mirror at a time is expected - unfortunately, this is a limitation of the spectator camera and is documented in the Readme. The freezing issue also appears to happen with the regular mirrors in A&J and I suspect is a limitation of the spectator camera frame at high resolutions. I will file a bug for the latter, and make sure that the former is also included in the Marketplace description!

zfox23:
People will want a Mirror blueprint, so we may revisit this code later - but this PR can be closed, since it isn't ready for merge in its current state and is unlikely that Liv'll work on it later :)

PR #236, opened 2018-12-19
Fix for plate pieces appearing in the wrong place

MiladNazeri:
The plate pieces were moving to the origin.  

Not sure exactly what was causing this, but I did notice some different behavior with the happy kiosk and its children.  It could be a similar issue local vs world and parent/child relationships. 

Either way, setting the position before the plates were made visible seemed to do the trick and that is what this pr is mainly taking care of.


salinacsoto:
Fix for plate pieces appearing in the wrong place - PASS
Tested plates in theSpot and the plates' broken pieces no longer spawn at origin point of the domain.  

QA Approved.

PR #237, opened 2018-12-19
New Year's Ball Drop

RebeccaStankus:
Testing edition--Ball will drop every 5 minutes.
TEST PLAN
-Get a time at least 2 minutes in the future and convert to GMT. Enter this time in the name of the countdown text entity and then reload the server script.
-Timer should show correct amount of minutes and seconds until the time and should be counting down
-At 1 minute, the ball should begin dropping and clock ticking sounds are heard.
-At 0, timer says "2019!", ball explodes with sound, fireworks with sound begin in random places in the sky, music plays, inner star is spinning inside of ball when it explodes and lights are rotating around it.
-After song finishes, fireworks stop, ball moves back to initial position and is a sphere again (not exploded), city list changes to next cities (or "HAPPY NEW YEAR EVERYONE!" if the list of cities is finished), timer is counting down again


DomainContent/NewYearBallDrop/ballDrop.js
```diff
@@ -0,0 +1,250 @@
...
+    var TICK = SoundCache.getSound(Script.resolvePath("assets/audio/tick.wav"));
+    var TOCK = SoundCache.getSound(Script.resolvePath("assets/audio/tock.wav"));
+    var HOURS_PER_DAY = 24;
+    var MAX_NUMBER_OF_DROPS = 27;
+
+    var injector;
+    var ballDropInterval;
+    var currentPosition = { x: -29.8177, y: 6.8299, z: -108.6157 };
+    var confetti;
+    var ball = "{6414271b-f0f4-4a62-b196-ac83dfd95253}";
```
MiladNazeri: I'm not sure if you should be doing this because if the domain resets, all of those ids are going to change.  This should probably be done with something more like:
var entityID = Entities.findEntitiesByName("{Entity-Name}", MyAvatar.position, 10, false);

DomainContent/NewYearBallDrop/ballDrop.js
```diff
@@ -0,0 +1,250 @@
...
+            targetDate = new Date(targetName.substring(TARGET_TIME_INDEX, (targetName.length)));
+            targetTime = targetDate.getTime();
+            targetYear = targetDate.getFullYear();
+            targetDay = targetDate.getDate();
+            targetHour = targetDate.getHours();
+            target5Minutes = targetDate.getMinutes();
+        },
+
+        synchronize: function() {
+            var currentTime = new Date().getTime();
```
MiladNazeri: you can also just do Date.now() for this

salinacsoto:
Tested and QA Approved :)

PR #239, opened 2018-12-26
New year ball drop

RebeccaStankus:
Testing edition--Ball will drop every 5 minutes.
TEST PLAN
-Get a time at least 2 minutes in the future and convert to GMT. Enter this time in the name of the countdown text entity and then reload the server script.
-Timer should show correct amount of minutes and seconds until the time and should be counting down
-At 1 minute, the ball should begin dropping and clock ticking sounds are heard.
-At 0, timer says "2019!", ball explodes with sound, fireworks with sound begin in random places in the sky, music plays, inner star is spinning inside of ball when it explodes and lights are rotating around it.
-After song finishes, fireworks stop, ball moves back to initial position and is a sphere again (not exploded), city list changes to next cities (or "HAPPY NEW YEAR EVERYONE!" if the list of cities is finished), timer is counting down again
-Check that noisemakers work and disappear after setting them down


zfox23:
Dropping this PR until we must revisit it - the code is not yet in a place to merge.

PR #242, opened 2019-01-03
Added Fix for grabbing textures

MiladNazeri:
This PR fixes how buttons were grabbing textures so that baking it shouldn't break it. 

Test plan should be simple, just make sure buttons have texture for off and on before baking and after baking.


salinacsoto:
QA Tested and Approved

PR #265, opened 2019-01-21
Update threads length when changing scale

luiscuenca:
This PR fix a scaling bug on flow.js. 

https://highfidelity.manuscript.com/f/cases/20492/flow-js-needs-to-detect-and-reset-after-avatar-scale-change-or-flow-bones-act-funny

Now the length of threads are also scaled. It also add some scaling to the delta parameter to avoid big changes on physics behavior when adopting extreme scale values.

### DIFF
--- a/Utilities/HairFlow/flow.js
+++ b/Utilities/HairFlow/flow.js
@@ -20,70 +20,70 @@ Script.include(Script.resolvePath("./VectorMath.js"));
     var SHOW_DEBUG_SHAPES = false;
     var SHOW_SOLID_SHAPES = false;
     var SHOW_DUMMY_JOINTS = false;
-    var USE_COLLISIONS = false;
-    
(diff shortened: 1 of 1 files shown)


Utilities/HairFlow/flow.js
```diff
@@ -20,70 +20,70 @@ Script.include(Script.resolvePath("./VectorMath.js"));
     var SHOW_DEBUG_SHAPES = false;
     var SHOW_SOLID_SHAPES = false;
     var SHOW_DUMMY_JOINTS = false;
-    var USE_COLLISIONS = false;
-    
+    var USE_COLLISIONS = true;
```
zfox23: making sure - did you mean to switch this bool?

luiscuenca: Yeah, I though it was better to activate collisions by default. I also changed some default params.

PR #270, opened 2019-01-25
Fix/party ball certified

MiladNazeri:
Added a fix for the dancers not being removed if a ball explodes and another is rezzed before it's over


MiladNazeri:
That was supposed to be ignored, it's the local mp json I use that is based on the relative path json

marketplaceItems/partyBall/entityResources/entityServerScripts/partyBall_server.js
```diff
@@ -295,7 +295,16 @@
 
     // Clear the explodeTimer if there is one
     function unload() {
-        explodeTimer && Script.clearInterval(explodeTimer);
+        if (explodeTimer) {
+            Script.clearInterval(explodeTimer);
+            Music.stop();
```
zfox23: should all of these functions get executed on unload even if `explodeTimer` is falsey?

MiladNazeri: hmm, I don't think so because the counterpart to those functions are only invoked to create entities if there is an explodeTimer created.

PR #297, opened 2019-02-16
Cleanup old IntensityEntities on startup/shutdown

zfox23:
This is necessary because sometimes High Fidelity crashes.


AndySeattle:
"Clean up crash claps" cleaned up!

PR #314, opened 2019-03-11
Bingo Boss App: Add a confirmation popup for buttons that start a new round

zfox23:
https://hifi-content.s3.amazonaws.com/Experiences/Releases/events/bingo/2019-03-11_11-16-00/bingoBossApp/bingoBossApp_app.js


salinacsoto:
Tested in both VR and Desktop and works. QA approved.

PR #316, opened 2019-03-12
Bingo: Add an exporter tool to make All Players Win prize-giving easier

zfox23:
This is a pretty HACKY way of accomplishing our goals, but it'll work! I need more backend experience...

### DIFF
--- a/events/bingo/webApps/bingo.js
+++ b/events/bingo/webApps/bingo.js
@@ -18,6 +18,7 @@
 var http = require('http');
 var url = require('url');
 var dbInfo = require('./dbInfo.json');
+var parseQueryString = require('querystring');
 
 // Creates a new table in the Bingo DB.
 var currentTableName;
@@ -40,7 +41,7 @@ function createNewTable(newTablePrefix, response) {
                 text: "Could not create new table! " + error
             };
     
-            response.statusCode = 200;
+            response.statusCode = 500;
             response.setHeader('Content-Type', 'application/json');
             return response.end(JSON.stringify(responseObject));
         }
@@ -124,7 +125,7 @@ function addNewPlayer(username, response) {
                 text: "Error adding new player!"
             };
 
-            response.statusCode = 200;
+            response.statusCode = 500;
             response.setHeader('Content-Type', 'application/json');
             return response.end(JSON.stringify(responseObject));
         }
@@ -178,7 +179,7 @@ function handleGetRequest(request, response) {
                         text: error
                     };
     
-                    response.statusCode = 200;
+                    response.statusCode = 500;
                     response.setHeader('Content-Type', 'application/json');
                     return response.end(JSON.stringify(responseObject));
                 }
@@ -212,7 +213,7 @@ function handleGetRequest(request, response) {
                 status: "error"
             };
 
-            response.statusCode = 200;
+            response.statusCode = 500;
             response.setHeader('Content-Type', 'application/json');
             return response.end(JSON.stringify(responseObject));
         }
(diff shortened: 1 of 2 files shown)


AndySeattle:
@zfox23 trying to test this PR, i am not seeing any instructions how? can you ping me with some instructions of what i should run for the exporter tool? :)

events/bingo/webApps/bingo.js
```diff
@@ -348,16 +379,127 @@ function replaceCalledNumbers(newCalledNumbers, response) {
...
+            status: "error",
+            text: "Not authorized."
+        };
+    
+        response.statusCode = 401;
+        response.setHeader('Content-Type', 'application/json');
+        return response.end(JSON.stringify(responseObject));    
+    }
+
+    var query = `SHOW TABLES LIKE '${formData.prefix}_%'`;
```
MiladNazeri: I looked this one up btw 
_ = match one char
% = match any number of chars
%cat% = blackcattail
so I don't think you need the _ 
 
​

events/bingo/webApps/bingo.js
```diff
@@ -348,16 +379,127 @@ function replaceCalledNumbers(newCalledNumbers, response) {
...
+
+        var allResults = [];
+
+        for (var i = 0; i < results.length; i++) {
+            var currentTableName;
+            for (var key in results[i]) {
+                currentTableName = results[i][key];
+            }
+            
+            query = `SELECT username,prizeWon FROM \`${currentTableName}\` WHERE username != 'BINGO BOSS'`;
```
MiladNazeri: I think this should work if you want to do one query :
 
query = `SELECT username,prizeWon FROM \`${currentTableName}\` WHERE username != 'BINGO BOSS'`;
var unionQuery = []
for (var i = 0; i < results.length; i++) {
var currentTableName;
for (var key in results[i]) {
currentTableName = results[i][key];
unionQuery.push(
`SELECT username, prizeWon FROM `${currentTableName}`
}
unionQuery = unionQuery.join(' UNION ALL SELECT ')
var finalQuery = `
SELECT username,prizeWon FROM(${unionQuery}) as B WHERE B.username != 'BINGO BOSS'
​
Might need to vamp a little, but that basic query string worked when trying in the db

MiladNazeri:
Really cool stuff, added some possible SQL changes from what we talked about yesterday

events/bingo/webApps/bingo.js
zfox23: let's say i have two of the same username in two different tables. will union make the output object contain only one of that username? or will the output object contain both?

events/bingo/webApps/bingo.js
zfox23: you're right that i don't NEED the semicolon, but it is more completely correct. table names are formatted `<prefix>_<timestamp>`

events/bingo/webApps/bingo.js
MiladNazeri: This just combines the tables together.   I checked manually joining 3 individual tables and the results came back to 125.  Those 3 tables had 128 listed, but the extra 3 were the bingo boss.

zfox23: fantastic! then it sounds like you're onto something. how can i then re-attribute the union'd data to their original table names?

MiladNazeri: Also adding here in case you need it: SELECT username,prizeWon, TABLE_NAME FROM 
(SELECT username,prizeWon, 'table1' as TABLE_NAME FROM `table1` 
     UNION ALL SELECT username,prizeWon, 'table2' as TABLE_NAME FROM `table2` 
     UNION ALL SELECT username,prizeWon, 'table3' as TABLE_NAME FROM `table3`) as B WHERE B.username != 'BINGO BOSS'

zfox23: copy/paste from our Slack convo:

you can make generated columns:
```SELECT username,prizeWon, TABLE_NAME FROM 
(SELECT username,prizeWon, 'table1' as TABLE_NAME FROM `feb28_1551295337686` 
     UNION ALL SELECT username,prizeWon, 'table2' as TABLE_NAME FROM `feb28_1551394893121` 
     UNION ALL SELECT username,prizeWon, 'table3' as TABLE_NAME FROM `feb28_1551376771649`) as B WHERE B.username != 'BINGO BOSS'```

this is perfect! i'll have to revamp this code. we can probably use it as-is for tomorrow's bingo, but i'd like to fix it before merging to master.

zfox23:
closing to revamp the tool

PR #327, opened 2019-03-20
"Deploy New Scripts" app!

zfox23:
See the README for usage instructions:
https://github.com/zfox23/hifi-content/tree/deployNewScripts/usefulUtilities/deployNewScripts


usefulUtilities/deployNewScripts/README.md
```diff
@@ -0,0 +1,24 @@
+# Deploy New Scripts App
+This app lets you quickly replace old client and server scripts on entities with new scripts.
+
+# Usage Instructions
```
RebeccaStankus: Should we mention that they may need to perform the swap in multiple places to 'find' all the entities that need to change? Is this even still necessary?

zfox23: I don't think this is a necessary thing to add to the README, although I do understand that what you're saying is true. If users of this app don't understand how to use it, I'll have to change something later (whether that's the app or the README).

RebeccaStankus:
When testing this on content for Mexico, the app was unable to change or even find the candles I tried to change. I copied the script directly from the create menu, so it had to be a match. Andy also got the same results. I was able to change all the food items, but it did not lock them again after.

zfox23:
QA approved as per @RebeccaStankus

PR #340, opened 2019-04-02
User Location Banner

zfox23:
Allows a user to rez a "banner" entity which displays their location.

https://hifi-content.s3.amazonaws.com/Experiences/Releases/usefulUtilities/userLocation/2019-04-02_09-52-00/userLocation.json
^ Edit the `userData`.

I have to add the "request" module to this folder; the entity script can't load the request module from disk.

### DIFF
--- a/usefulUtilities/userLocation/userLocation.js
+++ b/usefulUtilities/userLocation/userLocation.js
@@ -11,6 +11,11 @@
 
 
 (function () {
+    var request = Script.require("./modules/request.js").request;
+
+    var _this;
+    var REFRESH_LOCATION_INTERVAL_MS = 30000;
+    var USER_API_BASE_URL = Account.metaverseServerURL + "/api/v1/users?filter=connections&search=";
     var TEXT_LOCAL_ENTITY_PROPS = {
         "type": "Text",
         // eslint-disable-next-line no-magic-numbers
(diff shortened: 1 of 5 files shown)


salinacsoto:
bitmoji

usefulUtilities/userLocation/userLocation.js
```diff
@@ -0,0 +1,208 @@
...
+//  Copyright 2019 High Fidelity, Inc.
+//
+//  See accompanying README.md for usage instructions.
+// 
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+
+
+(function () {
+    var request = Script.require("./modules/request.js").request;
```
MiladNazeri: Is there a reason why we can't access request in client/server?  I think we are adding a lot of the same module over and over again.  Lol do a search for request in the hifi-content code base.

usefulUtilities/userLocation/userLocation.js
```diff
@@ -0,0 +1,208 @@
...
+    var USER_API_BASE_URL = Account.metaverseServerURL + "/api/v1/users?filter=connections&search=";
+    var TEXT_LOCAL_ENTITY_PROPS = {
+        "type": "Text",
+        // eslint-disable-next-line no-magic-numbers
+        "localPosition": [0, 0, 0.05],
+        // eslint-disable-next-line no-magic-numbers
+        "localRotation": Quat.fromVec3Degrees([0, 0, 0]),
+        "lineHeight": 0.08,
+        // eslint-disable-next-line no-magic-numbers
+        "dimensions": [1.5, 0.2, 0.05],
```
MiladNazeri: Ok I feel better about using arrays for vecs after seeing this, but taking away that it just depends on the context and if Milad can read it.

usefulUtilities/userLocation/userLocation.js
```diff
@@ -0,0 +1,208 @@
...
+        "script": Script.resolvePath("userLocationTextLocalEntity.js?6")
+    };
+
+    var UserLocationBanner = function() {
+        _this = this;
+    };
+
+    UserLocationBanner.prototype = {
+        preload: function(entityID) {
+            _this.entityID = entityID;
```
MiladNazeri: I like your use of consts at the top, but properties in the constructor itself.   Stealing.

usefulUtilities/userLocation/userLocation.js
```diff
@@ -0,0 +1,208 @@
...
+            _this.targetUsername = false;
+            _this.targetDisplayName = false;
+            _this.refreshLocationInterval = false;
+            _this.textLocalEntity = false;
+            _this.canTeleportToTargetUser = false;
+    
+            var properties = Entities.getEntityProperties(entityID, ["userData"]);
```
MiladNazeri: If the only property you are getting is userdata, then maybe just save yourself a var

usefulUtilities/userLocation/userLocation.js
```diff
@@ -0,0 +1,208 @@
...
+                }
+    
+                if (userData.targetDisplayName && userData.targetDisplayName.length > 0) {
+                    _this.targetDisplayName = userData.targetDisplayName;
+                } else {
+                    console.log("Please specify `targetDisplayName` inside this entity's `userData`!");
+                    return;
+                }
+            } else {
+                console.log("Please specify this entity's `userData`! See README.md for instructions.");
```
MiladNazeri: I wish we could get feedback like this not in the console.  Maybe we can use notifications for these, or something else that doesn't depend on the console to see why there was an error.

usefulUtilities/userLocation/userLocation.js
zfox23: discussed in Slack!

usefulUtilities/userLocation/userLocation.js
zfox23: 👍

usefulUtilities/userLocation/userLocation.js
zfox23: 👍

usefulUtilities/userLocation/userLocation.js
zfox23: this feels slightly cleaner to me personally!

zfox23: but good point :)

usefulUtilities/userLocation/userLocation.js
zfox23: i agree...notifications seem like the right move, if our notification system were better. i'll think about this for future projects. great idea

PR #350, opened 2019-04-06
Kayla cam update

MiladNazeri:


Prototyping/Kayla-Camera/webApp/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "miladmessage",
```
zfox23: please change these fields and regenerate package-lock

Prototyping/Kayla-Camera/webApp/dbClass.js
```diff
@@ -0,0 +1,39 @@
+// Wrapper for DB calls
```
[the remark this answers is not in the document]
MiladNazeri: 1. Promisifying dB calls
2. Removing business logic from the routing calls 
3. Ensuring error handling and proper opening and closing the dB connection

I wanted the routing section to be clean and focus as much as possible on what is needed for the routes.

Prototyping/Kayla-Camera/webApp/kaylaCamRouter.js
```diff
@@ -0,0 +1,111 @@
...
+            res.setHeader('Content-Type', 'application/json');
+            return res.end(JSON.stringify(responseObject));
+        }))
+})
+
+
+router.post('/settings', (req, res) => {
+    let query = `
+        INSERT INTO \`settings\` (config_name, settings)
+        VALUES ('${req.body.configName}', '${JSON.stringify(req.body)}')`
```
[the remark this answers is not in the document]
MiladNazeri: Are you referring to the settings table?  That is the only thing I see escaped and I did after the last cr of your pr.

Prototyping/Kayla-Camera/webApp/package.json
MiladNazeri: Nice catch!  Will do.

Prototyping/Kayla-Camera/webApp/kaylaCamRouter.js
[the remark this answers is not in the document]
zfox23: yeah, the settings table. i don't see where it's escaped here, can you point me to that please?

Prototyping/Kayla-Camera/webApp/dbClass.js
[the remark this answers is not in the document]
zfox23: cool. this is damn clean, and i don't see why it _should_ be any other way! cool stuff.

Prototyping/Kayla-Camera/webApp/kaylaCamRouter.js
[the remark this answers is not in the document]
MiladNazeri: I want to make sure we are seeing the same thing, this is what I have for the INSERT INTO part
image

And this is what I saw in your hifi ping
image

Prototyping/Kayla-Camera/webApp/kaylaCamRouter.js
```diff
@@ -49,7 +50,7 @@ router.get('/settings', (req, res) => {
 router.post('/settings', (req, res) => {
     let query = `
         INSERT INTO \`settings\` (config_name, settings)
-        VALUES ('${req.body.configName}', '${JSON.stringify(req.body)}')`
+        VALUES ('${req.body.configName}', '${mysql.escape(req.body)}')`
```
zfox23: doesn't `req.body.configName` also have to be escaped?

also, does this work? i thought `escape` was a method on a `connection` object?

MiladNazeri: These all call the same thing - In order to avoid SQL Injection attacks, you should always escape any user provided data before using it inside a SQL query. You can do so using the mysql.escape(), connection.escape() or pool.escape() methods:  

Maybe I'm a little confused, wouldn't that mean everything we are doing needs to be calling this?   For example in hifiPing:
```    var query = `REPLACE INTO \`subscriptions\` (username, subscription)
        VALUES ('${body.username}', '${JSON.stringify(body.subscription)}')`;```

zfox23: nice, ok, cool.

as for your second paragraph: yes, any user-submitted data that is going to be part of a MySQL query should be escaped. so yes, we (i, in this example) should really go back and update old code to prevent SQL Injection. it's a very low-priority issue right now, though.

MiladNazeri: I gotcha though, definitely more important for user-submitted values.  Will fix this.

PR #355, opened 2019-04-09
New statusIndicator HTML and API updates

zfox23:
- Changed the "GET DATA" API to return JSON instead of HTML
- Updated styling of HTML

### DIFF
--- a/usefulUtilities/statusIndicator/webApp/app/statusIndicator_webApp.js
+++ b/usefulUtilities/statusIndicator/webApp/app/statusIndicator_webApp.js
@@ -174,6 +174,25 @@ function updateEmployee(updates, response) {
 }
 
 
+function formatMemberData(singleResultObject) {
+    var location;
+    if (singleResultObject.status === "busy") {
+        location = "hidden";
+    } else {
+        location = singleResultObject.location || "unknown";
+    }
+
+    var displayName = singleResultObject.displayName;
+    var status = singleResultObject.status;
+
+    return {
+        "displayName": displayName,
+        "status": status,
+        "location": location
+    };
+}
+
+
 // Get all employees
 // Return tables organized by Team names with all employee information
 function getAllEmployees(response) {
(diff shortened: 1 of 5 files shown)


salinacsoto:
bitmoji

usefulUtilities/statusIndicator/webApp/app/statusIndicator_webApp.js
```diff
@@ -192,43 +211,35 @@ function getAllEmployees(response) {
...
+            "status": "success",
+            "teams": []
+        };
 
+        var currentTeamObject = {};
+        var currentTeamName = false;
         for (var i = 0; i < results.length; i++) {
-            if (teamName !== results[i].teamName) {
-                // different team name
+            if (currentTeamName !== results[i].teamName) {
```
MiladNazeri: Making sure I understand what is happening here.   currentTeamName starts out always being false, then if there is a team name it gets replaced.  Is currentTeamName being false handled else where(the tbd?)

usefulUtilities/statusIndicator/webApp/app/statusIndicator_webApp.js
```diff
@@ -192,43 +211,35 @@ function getAllEmployees(response) {
...
-    <td width="40%">${results[i].displayName}</td>
-    <td width="40%">${results[i].status}</td>
-    <td width="20%">${location}</td>
-</tr>
-            `;
+            currentTeamObject.members.push(formatMemberData(results[i]));
         }
-        responseHTML += `</table></div>`;
+        // Push the last one.
+        responseObject.teams.push(currentTeamObject);
```
MiladNazeri: This must be the false one then?

MiladNazeri:
Didn't see anything major, more just making sure I understand a section.

usefulUtilities/statusIndicator/webApp/app/statusIndicator_webApp.js
zfox23: right now, currentTeamName can't be false (i.e. there's always a value returned from the DB, even if it's "TBD"). so, there's no need to handle that case right now. however, you're totally right to point this out; i could definitely see a need in the future to handle that case.

usefulUtilities/statusIndicator/webApp/app/statusIndicator_webApp.js
zfox23: i'm not quite sure what you mean here. are you saying "this must be the object in which the team name is false" ? if so, no, that's not true, because the team name is never false.

MiladNazeri: Gotcha, was just not sure why that was pushed separately outside of the for loop?

zfox23: <discussed offline>

salinacsoto:
[Indicator](https://trello.com/c/3nJt8ndo/103-indicator)

PR #377, opened 2019-04-30
Status indicator update unknown to online

robin-k-wilson:
Unknown location will be seen as "online" on board.


AndySeattle:
Approving this one, looked at code, just a value swap on what is displayed. push push push

PR #389, opened 2019-05-15
App loader: quick fix

zfox23:
Fixed a bug where a duplicate copy of an already-running script would erroneously be loaded.


zfox23:
I have no idea why GitHub is saying that I changed the whole JS file - I didn't? I only changed a few lines in preload.

But, the script is small, and a full CR pass would be welcomed anyway.

PR #392, opened 2019-05-17
Sit bugs

RebeccaStankus:


usefulUtilities/sitV2/appResources/appData/sitServerV2.js
```diff
@@ -157,13 +156,44 @@
...
+            canSitZones.forEach(function(canSitZone) {
+                Entities.deleteEntity(canSitZone);
+            });
+        }
+    }
+
+    /* Set a random timeout with exponential growth to check for extra zones */
+    var ONE_SECOND_MS = 1000;
+    var ONE_TENTH_SECOND_MS = 100;
+    var EXPONENT = 2;
```
zfox23: this isn't an exponent in this implementation, it's more like a multiplier. please change the name of this variable to something like `MULTIPLIER`.

usefulUtilities/sitV2/appResources/appData/sitServerV2.js
```diff
@@ -157,13 +156,44 @@
...
+    var timeUntilNextSound = Math.floor((Math.random() * ONE_SECOND_MS) + ONE_TENTH_SECOND_MS);
```
zfox23: I _think_ what you want here is a random value between 100ms and 1000ms, right? If so, the code below will do that. Is that what you wanted?
```suggestion
    var timeUntilNextSound = Math.floor(Math.random() * (ONE_SECOND_MS - ONE_TENTH_SECOND_MS)) + ONE_TENTH_SECOND_MS;
```

usefulUtilities/sitV2/appResources/appData/sitServerV2.js
```diff
@@ -157,13 +156,44 @@
...
+    /* Check children of the sit cube to make sure there is only one zone. If more are found, delete them */
+    function checkForExtraZones() {
+        var canSitZones = [];
+        Entities.getChildrenIDs(_this.entityID).forEach(function(childOfSitCube) {
+            var name = Entities.getEntityProperties(childOfSitCube, 'name').name;
+            if (name && name.indexOf("canSitZone") > -1) {
+                canSitZones.push(childOfSitCube);
+            }
+        });
+        if (canSitZones.length > 1) {
```
zfox23: I'm confused about what this is doing. Why can't you do the `Entities.deleteEntity()` call in the `forEach` above? Why do you need an extra `canSitZones` list? Why aren't you referencing `_this.canSitZoneID` to avoid deleting that one?

usefulUtilities/sitV2/appResources/appData/sitServerV2.js
RebeccaStankus: I think it is fine if it gets a random number with a range of 1000ms beginning at100ms. Is there something wrong with the way it was done?

usefulUtilities/sitV2/appResources/appData/sitServerV2.js
RebeccaStankus: That would be better...changing it.

usefulUtilities/sitV2/appResources/appData/sitServerV2.js
```diff
@@ -157,13 +156,39 @@
     }
 
     // #endregion CAN SIT ZONE
+
+    /* Check children of the sit cube to make sure there is only one zone. If more are found, delete them */
+    function checkForExtraZones() {
+        Entities.getChildrenIDs(_this.entityID).forEach(function(childOfSitCube) {
+            var name = Entities.getEntityProperties(childOfSitCube, 'name').name;
+            if (name && name.indexOf("canSitZone") > -1) {
+                if (childOfSitCube !== _this.canSitZoneID) {
```
zfox23: please merge this `if` with the one above

usefulUtilities/sitV2/appResources/appData/sitServerV2.js
zfox23: no, there's nothing wrong with the way you did this. this seems like a fine implementation to me, i just didn't realize what the spec was :)

also, i just now noticed that this var name is `timeUntilNextSound` - please rename it to something relevant here

PR #394, opened 2019-05-21
Whiteboard bugs

RebeccaStankus:


salinacsoto:
Error still happens in VR

zfox23:
@salinacsoto can you please post the exact logs you're seeing in VR? i'm surprised this same error happens in VR - unless it's happening with other whiteboards

PR #398, opened 2019-05-23
New Portal

zfox23:
Yayyy, it's a new portal script that accepts relative paths!

### DIFF
--- a/usefulUtilities/portal/README.md
+++ b/usefulUtilities/portal/README.md
@@ -3,9 +3,9 @@ When this script is attached to a zone, the zone serves as a portal, which can b
 
 ## Setup
 1. Rez a zone entity that you want to act as the portal.
-2. Add the below `userData` object to the zone entity's `userData`
-3. Add the `portal.js` entity script to the zone entity
-4. Update `userData` at any time to modify the portal's destination
+2. Add the below `userData` object to the zone entity's `userData`.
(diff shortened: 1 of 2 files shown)


usefulUtilities/portal/README.md
```diff
@@ -0,0 +1,19 @@
+# Portal
+When this script is attached to a zone, the zone serves as a portal, which can be used to transport avatars between domains and within domains.
+
+## Setup
+1. Rez a zone entity that you want to act as the portal.
```
MiladNazeri: remove the period in the end to match the others

usefulUtilities/portal/portal.js
```diff
@@ -0,0 +1,40 @@
...
+            var userData;
+
+            try {
+                userData = JSON.parse(properties.userData);
+            } catch (e) {
+                console.error("Error parsing userData: ", e);
+            }
+
+            if (userData) {
+                if (userData.destination) {
```
MiladNazeri: Do you need to do a check here to make sure it's a valid destination?  Not sure what happens if you put something wonky in.

zfox23: Nope, we have no way to validate that. We're depending on the script user to supply a valid destination.

usefulUtilities/portal/README.md
zfox23: all set

PR #402, opened 2019-05-28
Clock

RebeccaStankus:
https://highfidelity.atlassian.net/browse/BUGZ-300


usefulUtilities/clock/entityResources/clockServer.js
```diff
@@ -71,7 +71,9 @@
             var hours = Number(date.getHours()) + Number(_this.timezoneOffset);
             hours = (hours < 0) ? hours + HOURS_PER_DAY : hours;
             if (hours >= AM_HOURS) {
-                period = "pm " + _this.timezoneName;
+                if (hours < HOURS_PER_DAY) {
+                    period = "pm " + _this.timezoneName;
+                }
                 if (hours !== AM_HOURS) {
```
zfox23: this does not handle the case when `hours === 0` - if that happens, we need to make `hours = 12`

PR #405, opened 2019-06-05
BUGZ-417 +BUGZ-431  White board controller refactor

MiladNazeri:
https://highfidelity.atlassian.net/browse/BUGZ-431
https://highfidelity.atlassian.net/browse/BUGZ-417

### DIFF
--- a/usefulUtilities/airAndWhiteboardDraw/README.md
+++ b/usefulUtilities/airAndWhiteboardDraw/README.md
@@ -21,7 +21,7 @@ May 21, 2019
 - Fixed bug with paint sphere dimensions not found [JIRA 152](https://highfidelity.atlassian.net/browse/BUGZ-152)
 - Made material entity attached to paint sphere collisionless 
 
-SHA [](https://github.com/highfidelity/hifi-content/commits/90ebb0d)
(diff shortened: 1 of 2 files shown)


qlipdahlin:
While using Vive VR, left hand laser would not activate to pick colors or reset the board. No laser appeared and pointing hand at correct location did not change anything either. Right hand did nothing either.

salinacsoto:
It is very difficult to delete lines in desktop with the middle scroll button on my Windows PC

salinacsoto:
hifi-snap-by-skelina-on-2019-06-07_14-25-25
I am able to pull lines out from the board on Master Build 10542. I tried using just simplified UI instead of having both and I still could repro. On my Macbook Air I could not repro.

salinacsoto:
In the oculus rift s I am able to draw, erase and trigger the buttons on the whiteboard normally

MiladNazeri:
Ooh, I hope the vive isn't different for some reason.  That's going to be a
hard one for me to check on with only an oculus system.

Peter, can you try the vive setup on the normal whiteboards and let me know
if they work as expected? That would be helpful to know.

PR #406, opened 2019-06-05
Sit

RebeccaStankus:


.gitignore
```diff
@@ -41,3 +41,5 @@
 ehthumbs.db
 Thumbs.db
 
+dbInfo.json
```
zfox23: why is this here?

RebeccaStankus: I added it to my git ignore because it somehow got removed

PR #408, opened 2019-06-06
Sit

RebeccaStankus:
https://highfidelity.atlassian.net/browse/BUGZ-554

### DIFF
--- a/usefulUtilities/sitV2/appResources/appData/sitClientV2.js
+++ b/usefulUtilities/sitV2/appResources/appData/sitClientV2.js
@@ -338,11 +338,11 @@
                 "addAllOtherSittableOverlays",
                 AvatarList.getAvatarsInRange(_this.seatCenterPosition, CAN_SIT_M)
             );
+            Controller.actionEvent.disconnect(onActionEvent);
             // Enable movement again
             for (var i in DISABLED_DRIVE_KEYS_DURING_SIT) {
                 MyAvatar.enableDriveKey(DISABLED_DRIVE_KEYS_DURING_SIT[i]);
             }
-            Controller.actionEvent.disconnect(onActionEvent);
         }, WAIT_FOR_USER_TO_STAND_MS);
     }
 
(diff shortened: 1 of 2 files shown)


usefulUtilities/sitV2/appResources/appData/sitClientV2.js
```diff
@@ -370,16 +332,17 @@
...
                 AvatarList.getAvatarsInRange(_this.seatCenterPosition, CAN_SIT_M)
             );
-            // ENABLE JUMP DRIVE KEYS
-            for (var i in REMAPPED_DRIVE_KEYS_DURING_SIT) {
-                MyAvatar.enableDriveKey(REMAPPED_DRIVE_KEYS_DURING_SIT[i]);
+            // Enable movement again
+            for (var i in DISABLED_DRIVE_KEYS_DURING_SIT) {
+                MyAvatar.enableDriveKey(DISABLED_DRIVE_KEYS_DURING_SIT[i]);
             }
+            Controller.actionEvent.disconnect(onActionEvent);
```
zfox23: Don't you want to do this _before_ re-enabling the drive keys?

RebeccaStankus: I guess there is an extremely slim chance that the space bar could be hit again. I will move it.

usefulUtilities/sitV2/appResources/appData/sitClientV2.js
```diff
@@ -339,10 +342,11 @@
                 "addAllOtherSittableOverlays",
                 AvatarList.getAvatarsInRange(_this.seatCenterPosition, CAN_SIT_M)
             );
-            
-            // Enable movement again
-            for (var i in DISABLED_DRIVE_KEYS_DURING_SIT) {
-                MyAvatar.enableDriveKey(DISABLED_DRIVE_KEYS_DURING_SIT[i]);
+            if (settingsEntityID === _this.entityID) {
```
zfox23: this concerns me a bit but i know there will be bugs with this implementation, so let's get this out there first!

RebeccaStankus: yeah, there is def an issue around changing seats right now and I think that will need to change but it fixes the more urgent problem with seat floating.

PR #412, opened 2019-06-12
BUGZ 653 :: Calendar encodeURI fix

MiladNazeri:
https://highfidelity.atlassian.net/projects/BUGZ/issues/BUGZ-653?filter=reportedbyme


MiladNazeri:
I don't think it needs QA because it doesn't work without it, and this change makes it work as expected again.

PR #431, opened 2019-07-01
[Dev-96] - New way to open Cal App

MiladNazeri:
https://highfidelity.atlassian.net/browse/DEV-96

### DIFF
--- a/usefulUtilities/hiFiCalendar/entityScripts/clickForTheCalendarAppClientScript.js
+++ b/usefulUtilities/hiFiCalendar/entityScripts/clickForTheCalendarAppClientScript.js
@@ -38,7 +38,7 @@
     // if it is, stop the script, then reload it again
     var RELOAD_SCRIPT = true;
     function maybeOpenTheCalendarApp() {
-        if (checkIfCalendarAppIsRunning()) { 
+        if (!checkIfCalendarAppIsRunning()) { 
             ScriptDiscoveryService.loadScript(_this.calendarAppURL);
         } else {
             ScriptDiscoveryService.stopScript(_this.calendarAppURL, RELOAD_SCRIPT) ;
@@ -50,7 +50,6 @@
     function maybeCloseTheCalendarApp() {
         if (checkIfCalendarAppIsRunning()) {
             ScriptDiscoveryService.stopScript(_this.calendarAppURL);
-            _this.isCalendarAppRunning = false;
         }
     }
 
--- a/usefulUtilities/hiFiCalendar/entityScripts/clickForTheCalendarAppClientScript.js
+++ b/usefulUtilities/hiFiCalendar/entityScripts/clickForTheCalendarAppClientScript.js
@@ -1,0 +1,88 @@
+//  clickForTheCalendarAppClientScript.js
+//
+//  Created by Milad Nazeri on 7-1-2019
+//  Copyright 2019 High Fidelity, Inc.
+// 
+//  Distributed under the Apache License, Version 2.0.
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+
+
+(function() {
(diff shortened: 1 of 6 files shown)


usefulUtilities/hiFiCalendar/entityScripts/clickForTheCalendarAppClientScript.js
```diff
@@ -0,0 +1,88 @@
...
+        console.log("mousePressOnEntity!")
+        if (event.isPrimaryButton) {
+            maybeOpenTheCalendarApp();
+        }
+    }
+
+
+    ClickForTheCalendarApp.prototype = {
+        unload: unload,
+        mousePressOnEnitty: mousePressOnEnitty,
```
zfox23: typo!

usefulUtilities/hiFiCalendar/entityScripts/clickForTheCalendarAppClientScript.js
```diff
@@ -0,0 +1,85 @@
...
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+
+
+(function() {
+
+    var _this = null;
+    function ClickForTheCalendarApp() {
+        _this = this;
+        
+        _this.isCalendarAppRunning = false;
```
zfox23: why do you need this flag?

usefulUtilities/hiFiCalendar/entityScripts/showClickForTheCalendarAppClientScript.js
```diff
@@ -0,0 +1,95 @@
...
+//  See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+
+
+(function() {
+
+    var _this = null;
+    function ShouldShowClickForTheCalendarApp() {
+        _this = this;
+        
+        _this.canGetAndSetPrivateUserData = false;
```
zfox23: why do you need this flag? seems like you don't need it and it adds extra state

usefulUtilities/hiFiCalendar/entityScripts/showClickForTheCalendarAppClientScript.js
```diff
@@ -0,0 +1,95 @@
...
+        _this.canGetAndSetPrivateUserData = Entities.canGetAndSetPrivateUserData();
+        _this.maybeCreateClickForTheCalendarAppEntity();
+
+        Entities.canGetAndSetPrivateUserDataChanged.connect(_this.onCanGetAndSetPrivateUserDataChanged);
+    }
+
+
+    // if you can get and set private user data, then create the local entity 
+    function maybeCreateClickForTheCalendarAppEntity() {
+        if (_this.canGetAndSetPrivateUserData) {
```
zfox23: for example, you could (and should) replace this with `if (Entities.canGetAndSetPrivateUserData()) {`

One reason why you "should" is because (even though it's unlikely) the value of the flag could change between the time you set the flag and got here.

usefulUtilities/hiFiCalendar/entityScripts/clickForTheCalendarAppClientScript.js
MiladNazeri: I don't, good call.

usefulUtilities/hiFiCalendar/entityScripts/showClickForTheCalendarAppClientScript.js
MiladNazeri: Agreed, removed.

usefulUtilities/hiFiCalendar/entityScripts/showClickForTheCalendarAppClientScript.js
MiladNazeri: I think it just being cautious, but calling the function is definitely better and simpler.

PR #440, opened 2019-07-17
added a test file

MiladNazeri:
bla

### DIFF
--- a/test.js
+++ b/test.js
@@ -1,0 +1,3 @@
+are you cat?
+am I cat?
+are we cat?
--- a/test.js
+++ b/test.js
@@ -1,3 +1,6 @@
 are you cat?
 am I cat?
 are we cat?
+
+We are not cat.  
+We have no tail. 


test.js
```diff
@@ -0,0 +1,3 @@
+are you cat?
+am I cat?
```
RebeccaStankus: What kind of cat?

test.js
```diff
@@ -0,0 +1,3 @@
+are you cat?
```
RebeccaStankus: I am.

test.js
```diff
@@ -0,0 +1,3 @@
+are you cat?
+am I cat?
+are we cat?
```
RebeccaStankus: I would say so.

test.js
MiladNazeri: DOES IT MATTER???

test.js
MiladNazeri: Am I?

test.js
MiladNazeri: cool.  checks out.

PR #441, opened 2019-07-23
Sound Emitter v1.2

zfox23:
- Added the ability to set the `position` option of the Audio Injector inside the Sound Emitter's `userData` via the `positionOverride` key/value pair.
- Documented why one would want to attach `soundEmitter.js` as either an entity client script or an entity server script.

https://content.highfidelity.com/Experiences/Releases/usefulUtilities/soundEmitter/v1.2/soundEmitter.js

### DIFF
--- a/usefulUtilities/soundEmitter/soundEmitter.js
+++ b/usefulUtilities/soundEmitter/soundEmitter.js
@@ -1,6 +1,6 @@
 /* eslint-disable no-magic-numbers */
 //
-// soundEmitterServer.js
+// soundEmitter.js
 // 
 // Created by Zach Fox on 2019-07-05
 // Copyright High Fidelity 2019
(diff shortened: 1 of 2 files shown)


usefulUtilities/soundEmitter/soundEmitter.js
```diff
@@ -111,6 +111,13 @@
         },
```
MiladNazeri: Not a big one or blocker, but I noticed this whole file doesn't have very many comments.  Just bringing it to your attention in case you want to add them.

salinacsoto:
As a client script, when the audio is playing and you switch the looping from true to false and save the user data you have to refresh the script itself to reflect that change. You only need to do this for the true/false of the looping. If you change the volume or position mid way and save the userdata it automatically updates.

PR #446, opened 2019-08-02
Fix BUGZ-1147: Audio log spam when in serverless domain

zfox23:
[BUGZ-1147](https://highfidelity.atlassian.net/browse/BUGZ-1147)


salinacsoto:
image

PR #451, opened 2019-08-13
Fix a bunch of Sit bugs and make it more stable

zfox23:
Fixes Sit and makes it way more stable.

Fixes the following bugs:
https://highfidelity.atlassian.net/browse/BUGZ-1229
https://highfidelity.atlassian.net/browse/DEV-376
https://highfidelity.atlassian.net/browse/DEV-377

Also fixes "the race condition" by using async callbacks when switching seats.

Doesn't fix the broken "heartbeat" system - if a server heartbeat fails (i.e. a user disconnects while sitting), the Sittable UI won't re-appear on that seat for any other users unless they restart their client. This has been broken since Sit V1.


salinacsoto:
image

usefulUtilities/sit/sitClient.js
```diff
@@ -73,55 +73,58 @@
```
RebeccaStankus: We could limit the properties retrieved at line 63 to be more efficient. Also, now that we are able to append this script to any entity to make it a "chair", always placing the avatar at 1.1 times the height of the "chair" object may not always give a realistic seat height. For instance, with a tall tree stump, placing the avatar at 1.1 times the height of the stump could leave a large gap between the top of the stump and the avatar.

usefulUtilities/sit/sitClient.js
```diff
@@ -134,36 +137,37 @@
...
-                MyAvatar.scaleChanged.connect(_this.standUp);
-                MyAvatar.onLoadComplete.connect(_this.standUp);
-                location.hostChanged.connect(_this.standUp);
-                Script.scriptEnding.connect(_this.standUp);
-                MyAvatar.wentAway.connect(_this.standUpWrapper);
-                HMD.displayModeChanged.connect(_this.standUpWrapper);
-                _this.connectedSignals = true;
-            }
-        }, SIT_DELAY_MS);
+        var properties = Entities.getEntityProperties(_this.entityID);
```
RebeccaStankus: It looks like we only use rotation so can we limit the request to just that property?

usefulUtilities/sit/sitClient.js
```diff
@@ -229,78 +233,68 @@
...
 
-        Entities.callEntityServerMethod(
-            _this.entityID,
-            "addAllOtherSittableOverlays",
-            AvatarList.getAvatarsInRange(_this.seatCenterPosition, CAN_SIT_M)
-        );
-        createSittableUI();
-        stopUpdateInterval();
+            // RESET SETTINGS FOR THIS CHAIR
+            // Could have changed seats, keep avatar sitting if did not go through above procedure
```
RebeccaStankus: I think you can remove this comment now

usefulUtilities/sit/sitClient.js
```diff
@@ -446,57 +439,66 @@
         canClickOnModelToSit: false
     };
     function updateUserData() {
```
RebeccaStankus: missing empty line between functions

usefulUtilities/sit/sitClient.js
```diff
@@ -446,57 +439,66 @@
         canClickOnModelToSit: false
     };
```
RebeccaStankus: missing empty line between functions

usefulUtilities/sit/sitClient.js
```diff
@@ -446,57 +439,66 @@
...
     function updateUserData() {
-        var properties = Entities.getEntityProperties(_this.entityID);
-        try {
-            _this.userData = JSON.parse(properties.userData);
-        } catch (e) {
-            console.log("Issue parsing userData" + e);
-        }
+        var userData = Entities.getEntityProperties(_this.entityID).userData;
```
RebeccaStankus: One more place where we can limit the properties requested

usefulUtilities/sit/sitClient.js
```diff
@@ -446,57 +439,66 @@
...
 
-        if (!_this.userData || _this.userData.canClickOnModelToSit === undefined) {
+        if (!userData || _this.userData.canClickOnModelToSit) {
```
RebeccaStankus: Is this supposed to rewrite user data for canClickOnModelToSit to false every time?

usefulUtilities/sit/sitClient.js
```diff
@@ -446,57 +439,66 @@
...
-    function check() {
-        Entities.callEntityServerMethod(_this.entityID, "checkResolved");
+    function heartbeatRequest() {
+        Entities.callEntityServerMethod(_this.entityID, "heartbeatResponse");
     }
 
     
     // Can sit when clicking on chair when enabled via userData
+    var MAX_SIT_DISTANCE_M = 5;
+    EDIT_SETTING = "io.highfidelity.isEditing";
```
RebeccaStankus: missing "var" in the declaration

usefulUtilities/sit/sitClient.js
```diff
@@ -543,24 +543,25 @@
...
-            "deleteSittableUI",
-            "startSitDown",
-            "check"
+            "createClickToSitOverlay",
+            "deleteClickToSitOverlay",
+            "checkBeforeSitDown",
+            "heartbeatRequest",
+            "startSitDown"
         ],
         // Zone methods
```
RebeccaStankus: No more zones on sit cubes so we can remove the comment

usefulUtilities/sit/sitServer.js
```diff
@@ -0,0 +1,224 @@
+//
+// sitServer.js
+//
+// Created by Robin Wilson 1/17/2019
+//
+// Copyright 2017 High Fidelity, Inc.
```
RebeccaStankus: 2019

usefulUtilities/sit/sitServer.js
```diff
@@ -0,0 +1,224 @@
...
+//
+// Distributed under the Apache License, Version 2.0.
+// See the accompanying file LICENSE or http://www.apache.org/licenses/LICENSE-2.0.html
+// 
+// sitServer.js is the gate to allow an avatar to sit. Multiple avatars can not sit in the same chair.
+// Spawns the canSitZone per chair
```
RebeccaStankus: not accurate comment anymore since we removed the zones

usefulUtilities/sit/sitServer.js
```diff
@@ -0,0 +1,224 @@
...
+                    _this.entityID,
+                    "createClickToSitOverlay"
+                );
+            }
+        }
+    }
+
+
+    // Remotely callable
+    // Remove sittable local entities from every client in range passed in by sitClient
```
RebeccaStankus: not dependent on range anymore

usefulUtilities/sit/sitClient.js
zfox23: ACK

Regarding the 1.1x multiplier, you're right. We don't have a better way of doing that right now, and, so far, there haven't been any complaints about such an issue. I will add some text to the Known Issues section of the README about this.

usefulUtilities/sit/sitClient.js
zfox23: ACK

usefulUtilities/sit/sitClient.js
zfox23: ACK

zfox23: oh, this comment is almost correct, it's just slightly misleading. when the comment says "settings", it means "locked state". i'll update

zfox23: oh, unless you were only talking about the second line here, in which case yeah, i gotta remove that

usefulUtilities/sit/sitClient.js
zfox23: I think what's here is correct; `DEFAULT_SIT_USER_DATA_WITH_CUSTOM_SETTINGS` is not a function

usefulUtilities/sit/sitClient.js
zfox23: ACK

usefulUtilities/sit/sitClient.js
zfox23: No. This code is a bit of a mess, lemme fix it up

usefulUtilities/sit/sitClient.js
zfox23: ACK

usefulUtilities/sit/sitClient.js
zfox23: ack

usefulUtilities/sit/sitServer.js
zfox23: ACK

usefulUtilities/sit/sitServer.js
zfox23: ack

usefulUtilities/sit/sitServer.js
zfox23: ack

PR #462, opened 2019-09-16
Configurable courtyard

RebeccaStankus:
https://highfidelity.atlassian.net/browse/DEV-296


placeSpecificContent/hq/courtyard/README.md
```diff
@@ -0,0 +1,28 @@
...
+* Roundtable - (HiFi Utility) empty.js
+* Admin Only Zone - (HiFi Utility) bouncer.js
+
+## File List
+### Server Side
+* **controlPanelButton_client.js** attached to the privacy button entity.
+* **empty.js** attached to the chatterbox, table, shade, and door entities.
+
+### Client Side
+* **controlPanelButton_client.js** attached to the privacy button entity.
```
zfox23: copy/paste error? "privacy button entity"

placeSpecificContent/hq/courtyard/README.md
```diff
@@ -0,0 +1,28 @@
...
+    * Control Panel Button Roundtable - controlPanelButton_client.js
+    * Control Panel Button Courtyard - controlPanelButton_client.js
+* Courtyard - (HiFi Utility) empty.js
+    * Safety Zone - (HiFi Utility) bouncer.js
+* Roundtable - (HiFi Utility) empty.js
+* Admin Only Zone - (HiFi Utility) bouncer.js
+
+## File List
+### Server Side
+* **controlPanelButton_client.js** attached to the privacy button entity.
```
zfox23: copy/paste error? "privacy button entity"

placeSpecificContent/hq/courtyard/README.md
```diff
@@ -0,0 +1,28 @@
...
+* **empty.js** attached to the chatterbox, table, shade, and door entities.
```
zfox23: copy/paste error?

placeSpecificContent/hq/courtyard/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+### Client Side
+* **controlPanelButton_client.js** attached to the privacy button entity.
+* **bouncer.js** attached to the admin only zone around the control panel room and the safety check zone beneath the courtyard floor.
```
zfox23: Link to bouncer zone readme page on github please

placeSpecificContent/hq/courtyard/controlPanel/controlPanel_server.js
```diff
@@ -0,0 +1,406 @@
...
+                    roundtableInPlace = true;
+                } else {
+                    courtyardInPlace = true;
+                }
+                return;
+            }
+            
+            // If the entity has somehow been moved far away, slam it into lowered position to avoid hitting anyone
+            if (Math.abs(distanceToTarget) > MAX_DISTANCE_TO_TRANSITION_ENTITIES_M && 
+                Math.abs(distanceToTarget) < -MAX_DISTANCE_TO_TRANSITION_ENTITIES_M) {
```
zfox23: this will never be true, right? `Math.abs` will always return a positive number. why do you have to do `Math.abs(distanceToTarget)` ? why isn't `distanceToTarget` always positive, since it's a distance?

placeSpecificContent/hq/courtyard/controlPanel/controlPanel_server.js
```diff
@@ -0,0 +1,406 @@
...
+                currentPosition = entityID === roundtable ? roundtableLoweredPosition : courtyardLoweredPosition;
+                Entities.editEntity(entityID, {
+                    position: currentPosition
+                });
+            // If entity is very close to target position, slam into exact position
+            } else if (Math.abs(distanceToTarget) < MIN_DISTANCE_TO_TRANSITION_ENTITIES_M && 
+                Math.abs(distanceToTarget) > -MIN_DISTANCE_TO_TRANSITION_ENTITIES_M) {
```
zfox23: this will never be true, right? `Math.abs` will always return a positive number. why do you have to do `Math.abs(distanceToTarget)` ? why isn't `distanceToTarget` always positive, since it's a distance?

placeSpecificContent/hq/courtyard/controlPanel/controlPanel_server.js
```diff
@@ -0,0 +1,406 @@
...
+                        backgroundColor: BUTTON_INACTIVE_COLOR,
+                        textColor: BUTTON_TRANSITION_COLOR
+                    });
+                }
+            });
+
+            var soundPosition = Entities.getEntityProperties(courtyard, 'position').position;
+            _this.playSound(TRANSITION_SOUND, TRANSITION_VOLUME, soundPosition, false, true);
+            
+            if (!roundtableInPlace && roundtableVelocity) {
```
zfox23: when would `roundtableInPlace` ever be `true` here?
when would `roundtableVelocity` ever be falsey here?

placeSpecificContent/hq/courtyard/controlPanel/controlPanel_server.js
```diff
@@ -0,0 +1,406 @@
...
+                Entities.editEntity(roundtable, {
+                    velocity: { x: 0, y: roundtableVelocity, z: 0 }
+                });
+            }
+
+            if (!courtyardInPlace && courtyardVelocity) {
```
zfox23: when would `courtyardInPlace` ever be `true` here?
when would `courtyardVelocity` ever be falsey here?

placeSpecificContent/hq/courtyard/controlPanel/controlPanel_server.js
```diff
@@ -0,0 +1,406 @@
...
+                Entities.editEntity(courtyard, {
+                    velocity: { x: 0, y: courtyardVelocity, z: 0 }
+                });
+            }
+
+            if (checkInterval) {
```
zfox23: why wouldn't this be the first thing you do inside `pressButton`?

placeSpecificContent/hq/courtyard/controlPanel/controlPanel_server.js
```diff
@@ -0,0 +1,406 @@
...
+        },
+
+        /* Check each moveable entity and if it is still moving and close enough to its target position, slam it and clear 
+        its velocity. If both entities are in place, play the done sound and set the button color to active and stop the 
+        position checking interval.
+        */
+        checkEntityPositions: function(buttonID, targetCourtyardPosition, targetRoundtablePosition) {
+            if (!courtyardInPlace) {
+                var courtyardCurrentPosition = Entities.getEntityProperties(courtyard, 'position').position;
+                var courtyardDistanceToTarget = courtyardCurrentPosition.y.toFixed(FIXED_DECIMAL_PLACES) -
```
zfox23: - Why are you using `toFixed` here?
- Why not use `Vec3.distance()` here?
- This isn't a distance, since it could be negative. Did you intend for this to be a distance? If so, you'll need to wrap this in `Math.abs()` (but only if you're not going to use `Vec3.distance()`)

placeSpecificContent/hq/courtyard/controlPanel/controlPanel_server.js
```diff
@@ -0,0 +1,406 @@
...
+                var courtyardDistanceToTarget = courtyardCurrentPosition.y.toFixed(FIXED_DECIMAL_PLACES) - 
+                targetCourtyardPosition.y.toFixed(FIXED_DECIMAL_PLACES);
+                if ((courtyardVelocity < 0 && courtyardDistanceToTarget <= 0) || 
+                    (courtyardVelocity > 0 && courtyardDistanceToTarget >= 0)) {
+                    courtyardInPlace = true;
+                }
+                if (courtyardInPlace) {
+                    courtyardVelocity = { x: 0, y: 0, z: 0 };
+                    Entities.editEntity(courtyard, {
+                        velocity: { x: 0, y: 0, z: 0 },
```
zfox23:
```suggestion
                        velocity: courtyardVelocity,
```

placeSpecificContent/hq/courtyard/controlPanel/controlPanel_server.js
```diff
@@ -0,0 +1,406 @@
...
+                        position: targetCourtyardPosition
+                    });
+                }
+            }
+
+            if (!roundtableInPlace){
+                var roundtableCurrentPosition = Entities.getEntityProperties(roundtable, 'position').position;
+                var roundtableDistanceToTarget = roundtableCurrentPosition.y.toFixed(FIXED_DECIMAL_PLACES) -
```
zfox23: - Why are you using `toFixed` here?
- Why not use `Vec3.distance()` here?
- This isn't a distance, since it could be negative. Did you intend for this to be a distance? If so, you'll need to wrap this in `Math.abs()` (but only if you're not going to use `Vec3.distance()`)

placeSpecificContent/hq/courtyard/controlPanel/controlPanel_server.js
```diff
@@ -0,0 +1,406 @@
...
+                    currentPosition = {
+                        x: currentPosition.x,
+                        y: courtyardProperties.position.y - tableHeight * HALF,
+                        z: currentPosition.z
+                    };
+                    Entities.editEntity(roundtable, { position: currentPosition });
+                }
+            }
+            
+            var distanceToTarget = currentPosition.y.toFixed(FIXED_DECIMAL_PLACES) -
```
zfox23: - Why are you using `toFixed` here?
- Why not use `Vec3.distance()` here?
- This isn't a distance, since it could be negative. Did you intend for this to be a distance? If so, you'll need to wrap this in `Math.abs()` (but only if you're not going to use `Vec3.distance()`)

placeSpecificContent/hq/courtyard/controlPanel/controlPanel_server.js
```diff
@@ -0,0 +1,406 @@
...
+                        y: courtyardProperties.position.y - tableHeight * HALF,
+                        z: currentPosition.z
+                    };
+                    Entities.editEntity(roundtable, { position: currentPosition });
+                }
+            }
+            
+            var distanceToTarget = currentPosition.y.toFixed(FIXED_DECIMAL_PLACES) - 
+                targetPosition.y.toFixed(FIXED_DECIMAL_PLACES);
+            if (distanceToTarget.toFixed(FIXED_DECIMAL_PLACES) === 0 ||
```
zfox23: It's dangerous to compare mutable floating point values to 0 exactly. Please use some small epsilon for comparison instead of 0.

placeSpecificContent/hq/courtyard/controlPanel/controlPanel_server.js
```diff
@@ -0,0 +1,406 @@
...
+                    position: targetPosition
+                });
+                if (entityID === roundtable) {
+                    roundtableInPlace = true;
+                } else {
+                    courtyardInPlace = true;
+                }
+            }
+            // If entity is not aligned along x and z axes, slam into correct position on those axes
+            if (currentPosition.x.toFixed(FIXED_DECIMAL_PLACES) !== targetPosition.x.toFixed(FIXED_DECIMAL_PLACES) ||
```
zfox23: It's dangerous to compare mutable floating point values to other floating point values exactly. Please use some small epsilon for comparison instead of `===`.

placeSpecificContent/hq/courtyard/controlPanel/controlPanel_server.js
RebeccaStankus: I will remove the < comparison (it is left over piece of code that did not get removed) and rename the variable to not use the word 'distance', it is the displacement along the y axis from the current position point to the target position. I use the negative/positive aspect to determine which direction the entity needs to move and then set its velocity.

placeSpecificContent/hq/courtyard/controlPanel/controlPanel_server.js
```diff
@@ -0,0 +1,406 @@
...
+// See accompanying license file or http://apache.org/
+//
+
+(function() {
+    var HALF = 0.5;
+
+    var DATA_RETRIEVAL_TIMEOUT_MS = 500;
+    var PRELOAD_DATA_RETRIEVAL_MAX_ATTEMPTS = 10;
+
+    var DONE_SOUND = SoundCache.getSound(Script.resolvePath("resources/sounds/DONE_SOUND.mp3?1"));
```
MiladNazeri: I think we are trying to get away from cache busting so much

placeSpecificContent/hq/courtyard/controlPanel/controlPanel_server.js
```diff
@@ -0,0 +1,406 @@
...
+    var NUMBER_OF_BUTTONS = 3;
+    var BUTTON_TO_ENTITY_POSITIONING_MAP = {};
+
+    var FIXED_DECIMAL_PLACES = 2;
+    /* MAP WILL LOOK LIKE THIS BUT BUTTONS MAY NOT BE LISTED IN THIS ORDER: 
+
+    {
+        "{00000000-0000-0000-0000-000000000000}"(id of stage button): {
+            "courtyardPosition": { x: 0, y: 0, z: 0 } (position of courtyard when in the stage state)
+            "roundtablePosition": { x: 0, y: 0, z: 0 } (position of courtyard when in the stage state)
```
MiladNazeri: just a little confused on if what we are asking for is the position of courtyard on all of these based on the name of those object keys

placeSpecificContent/hq/courtyard/controlPanel/controlPanel_server.js
```diff
@@ -0,0 +1,406 @@
...
+            if (JSON.stringify(BUTTON_TO_ENTITY_POSITIONING_MAP) !== "{}" && roundtable && courtyard) {
+                ready = true;
+            } else if (preloadIntervalCount < PRELOAD_DATA_RETRIEVAL_MAX_ATTEMPTS) {
+                _this.setDataRequestTimeout();
+            }
+        },
+
+        /* Set a timeout for 500ms after which we will attempt to read entity data */
+        setDataRequestTimeout: function() {
+            dataRetrievalTimeout = Script .setTimeout(function() {
```
MiladNazeri: that says "Script .setTimeout" with extra space

placeSpecificContent/hq/courtyard/controlPanel/controlPanel_server.js
RebeccaStankus: Yes, we need the postion of both the courtyard and the roundtable at each these states: stage, courtyard, roundtable

placeSpecificContent/hq/courtyard/controlPanel/controlPanel_server.js
RebeccaStankus: OH, yeah, I just forgot to remove that for CR. Thanks!

placeSpecificContent/hq/courtyard/README.md
RebeccaStankus: yep, fixed.

placeSpecificContent/hq/courtyard/README.md
RebeccaStankus: yes

placeSpecificContent/hq/courtyard/README.md
RebeccaStankus: yes, fixed

placeSpecificContent/hq/courtyard/controlPanel/controlPanel_server.js
RebeccaStankus: Explained above.

placeSpecificContent/hq/courtyard/controlPanel/controlPanel_server.js
RebeccaStankus: If the roundtable does not need to move, 'roundtableInPlace' gets set at line 334 or line 356 in the setUpMovement function. The roundtableVelocity is also set in this function.

placeSpecificContent/hq/courtyard/controlPanel/controlPanel_server.js
RebeccaStankus: same as above for roundtable.

placeSpecificContent/hq/courtyard/controlPanel/controlPanel_server.js
RebeccaStankus: I think it should be moved up to the top of the function but still after the check for ready so we avoid an extra check if the script was unable to collect the data needed to handle state changes.

placeSpecificContent/hq/courtyard/controlPanel/controlPanel_server.js
RebeccaStankus: I covered this above. It will be 'displacement' along the y axis instead of 'distance'. The entity only has animated movement along one axis so I only check that axis.  I used toFixed in order to round the numbers for comparison as the final value of the entity's position on the y axis, when rounded, should be equal to the the rounded value of the target position on the y axis. This could be changed to use an epsilon value to get a comparison but I feel it accomplishes the same goal.

placeSpecificContent/hq/courtyard/controlPanel/controlPanel_server.js
RebeccaStankus: https://github.com/highfidelity/hifi-content/pull/462#discussion_r325268910

placeSpecificContent/hq/courtyard/controlPanel/controlPanel_server.js
RebeccaStankus: https://github.com/highfidelity/hifi-content/pull/462#discussion_r325268910

placeSpecificContent/hq/courtyard/controlPanel/controlPanel_server.js
RebeccaStankus: I was confused on toFixed vs rounding. If I change this to round to 2 decimal places will it accomplish the same thing as using an epsilon?

placeSpecificContent/hq/courtyard/controlPanel/controlPanel_server.js
RebeccaStankus: https://github.com/highfidelity/hifi-content/pull/462#discussion_r325276438

placeSpecificContent/hq/courtyard/controlPanel/controlPanel_server.js
```diff
@@ -0,0 +1,403 @@
...
+
+                var controlPanelProperties = Entities.getEntityProperties(_this.entityID, 'userData');
+                var controlPanelUserData = JSON.parse(controlPanelProperties.userData);
+
+                Entities.getChildrenIDs(_this.entityID).forEach(function(controlPanelChildID) {
+
+                    var name = Entities.getEntityProperties(controlPanelChildID, 'name').name;
+                    if (name.indexOf("Control Panel Button") > -1) {
+
+                        var stateName = name.substr(INDEX_WHERE_STATE_NAME_BEGINS).toLowerCase();
```
zfox23: Why can't you use `indexOf()` instead of relying on the entity name matching 100%?

RebeccaStankus: I can. I thought it would be better to be more exact in case anything else gets parented to the control panel at any point to ensure there are no mix-ups.

zfox23: ahhh, that makes sense. OK. i'm fine with this :)

PR #478, opened 2019-11-26
Smartboard v1.3: Attempt to fix DEV-2821 and DEV-2815

zfox23:
[DEV-2821](https://highfidelity.atlassian.net/browse/DEV-2821)
[DEV-2815](https://highfidelity.atlassian.net/browse/DEV-2815)

See tickets for test plan and JSON.

### DIFF
--- a/usefulUtilities/smartBoard/scripts/boardButtonClient.js
+++ b/usefulUtilities/smartBoard/scripts/boardButtonClient.js
@@ -12,22 +12,35 @@
 (function() {
     var DEBUG = false;
 
-    // ENTITY SIGNALS
-    function preload(entityID) {
-        if (DEBUG) {
-            console.log("boardButtonClient.js: " + entityID + ": `preload()`.");
-        }
-        _this.entityID = entityID;
-        var props = Entities.getEntityProperties(entityID, ["parentID"]);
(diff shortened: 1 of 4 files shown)


usefulUtilities/smartBoard/scripts/boardButtonClient.js
```diff
@@ -12,22 +12,35 @@
...
 
-        var children = Entities.getChildrenIDs(props.parentID);
-        
-        for (var i = 0; i < children.length; i++) {
-            var name = Entities.getEntityProperties(children[i], 'name').name;
-            if (name && name === "Smartboard Zone") {
-                _this.screenshareZoneID = children[i];
-                break;
+        var screenshareZoneIDRetries = 0;
+        while (!_this.screenshareZoneID) {
```
MiladNazeri: are you concerned with timing at all?  this while loop could be running very fast.

zfox23: Somewhat, yeah, but even if this does happen really fast, the other stuff that needs to happen before this is also happening very fast. If we still have a bug because of this code, we'll change this to a timer with short timeout.

